PR #289, opened 2021-12-09
Oy2 14080 bp

benjaminpaige:
moved the coreQuestions into the demo measure
broke out the questions into questions folder within measure

id like to see if we can use the useCustomRegister for all elements, if for nothing other than the types. Also we will really need to figure out a good naming pattern. Ill check with Anna to see if she can help with that.


13bfrancis (approved):
I love this

jdinh8124 (approved):

abelzer212 (approved):

PR #292, opened 2021-12-10
Sticky nav save thing

benjaminpaige:
Add sticky nav thing and save / submit buttons in layout.

https://qmacbis.atlassian.net/browse/OY2-14426


services/ui-src/src/components/index.tsx
```diff
@@ -1,14 +1,16 @@
+export * from "./Breadcrumbs";
```
abelzer212, line 1: I want it to be NON-alphabetical
As a side there's an ESLINT rule that does this automatically

abelzer212 (approved):

PR #313, opened 2021-12-20
Add TypeScript to Backend

13bfrancis:
This will add TypeScript to API


abelzer212 (approved):

services/app-api/handlers/measures/create.ts
```diff
@@ -3,21 +3,32 @@ import dynamoDb from "../../libs/dynamodb-lib";
 import { createCompoundKey } from "../dynamoUtils/createCompoundKey";
 
 export const createMeasure = handler(async (event, context) => {
+  if (!event.body) return; //throw
+  const body = JSON.parse(event.body);
+  if (!event.pathParameters) return; // throw error message
```
jdinh8124, line 8: The error messages going to be handled in the api branch?

services/app-api/handlers/measures/get.ts
```diff
@@ -4,14 +4,23 @@ import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpress
...
+    !event.pathParameters.coreSet
+  )
+    return; // throw error message
+
   const state = event.pathParameters.state;
   const year = event.pathParameters.year;
   const coreSet = event.pathParameters.coreSet;
 
+  console.log(state, year, coreSet);
   const params = {
```
jdinh8124, line 20: console log from our debugging earlier, not sure if you want that still

services/app-api/libs/debug-lib.ts
```diff
@@ -1,3 +1,5 @@
+//@ts-nocheck
+
 import util from "util";
 import AWS from "aws-sdk";
```
jdinh8124, line 5: wondering if we could blow this file up, considering it seemed like boilerplate

jdinh8124 (approved):
Thanks for getting this done! Going to help us a lot down the road

services/app-api/handlers/measures/get.ts
13bfrancis: I'll get that out. Good catch

services/app-api/libs/debug-lib.ts
13bfrancis: Another good catch. I'll see if it's used

services/app-api/handlers/coreSets/create.ts
```diff
@@ -37,13 +46,20 @@ export const createCoreSet = handler(async (event, context) => {
...
 });
 
-const createDependentMeasures = async (state, year, coreSet, type) => {
+const createDependentMeasures = async (
+  state: string,
+  year: number,
+  coreSet: string,
+  type: string
+) => {
+  if (year !== 2021 && year !== 2022) return;
```
benjaminpaige, line 61: Can we use the keys from the static measuresList for this?

13bfrancis: Oooh. Really good idea. We should definitely be able to do that. That will be much more dynamic too

abelzer212: No we can't the Keys are on the frontend, and this lives in AWS

services/app-api/handlers/measures/create.ts
13bfrancis: We'll handle that afterwards

services/app-api/libs/debug-lib.ts
13bfrancis: I'm gonna leave it for now

benjaminpaige (approved):

PR #314, opened 2021-12-20
Measures routing and measures wrapper

benjaminpaige:
Added dynamic routing for measures
Move types from demo measure into measure directory
Added MeasureWrapper comp


services/ui-src/src/measures/2021/AMM.tsx
```diff
@@ -0,0 +1,12 @@
...
+// This is where we will put the measures - the exported comp should be named as the measureId (without the trailing '-XX' )
+// All the measures defined in this directory will have all the props from the MeasureWrapper passed into it
+export const AMM = ({ name, year }: Measure.Props) => {
+  return (
+    <>
+      <p>{name}</p>
+      <p>{year}</p>
+    </>
+  );
+};
```
13bfrancis, line 12: I really like this as a convention. Makes things easy to follow

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <QMR.MeasureButtons
+            handleSave={methods.handleSubmit(handleSave)}
+            handleSubmit={methods.handleSubmit(handleSubmit)}
+            lastSavedText="Saved Moments Ago"
+          />
+        }
+      >
+        <form>
+          <CUI.Box maxW="7xl" as="section">
+            {cloneElement(measure, { name, year })}
```
13bfrancis, line 71: I like it!!

13bfrancis (approved):
This is good. It will bring in a lot of much needed organization to our frontend for the measures and the handling of them.

services/ui-src/src/measures/2021/index.tsx
```diff
@@ -0,0 +1,12 @@
+/*
+When importing a measure it should be a named import and added to the measures object below so that it routes correctly
+the key should be the measure id as a string (with '-XX' included) 
+*/
+
+import { AMM } from "./AMM";
```
abelzer212, line 6: Are we sure that the Adult/Child/Health Home variants are all identical besides name? 
So will we need to specify on import: AMM-AD vs AMM-CH vs AMM-HH?

benjaminpaige: I am 99% sure, but will confirm the extra 1% also maybe we can just drop the '-' and export from indexs. ill try that instead, that is prob better, good call.

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  this would include:
+  - validating the route params (state, coreset)
+  - querying the measure data
+  - defining the update method
+  - handing top level errors
+  - loading states (maybe?)
+
+  all of the methods defined here can be passed as props to every measure below
+  */
+
```
jdinh8124, line 31: Excited to use and consume this mesasureWrapper, it should make our lives easier

jdinh8124 (dismissed):

PR #320, opened 2021-12-21
Oy2 14511/14512/14514 fix

jdinh8124:


services/ui-src/src/views/CoreSet/index.tsx
```diff
@@ -31,17 +31,17 @@ enum coreSetQuestionsText {
...
+  // This is where a fetch for the measures would live and calculate progress completed
   const measures = [
     {
-      Type: "Adult",
-      title: "Antidepressant Medication Management",
-      abbr: "AMM-AD",
+      Type: "Health Home",
+      title: "Ambulatory Care: Emergency Department (ED) Visits",
+      abbr: "AMB-HH",
       path: "",
```
benjaminpaige, line 40:
```suggestion
      path: `/${state}/${year}/${coreSetId}/AMB-HH`,
```

benjaminpaige (commented):
looks great, can we just add paths to the example measures?

services/ui-src/src/views/CoreSet/index.tsx
jdinh8124: good idea!

benjaminpaige (approved):

PR #348, opened 2021-12-29
UI auth

abelzer212:
Bring out user authentication into its own separate functionality and added readonly context that should disable forms when not logged in as state user


services/ui-src/src/components/Layouts/StateLayout/index.tsx
```diff
@@ -32,7 +32,7 @@ export const StateLayout: React.FC<Props> = ({
     <>
       <CUI.Box
         position={sticky ? "fixed" : "relative"}
-        zIndex="2"
+        zIndex="1000"
```
benjaminpaige, line 35: why are we changing this? is there something with a higher index?

services/ui-src/src/hooks/authHooks/userProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+          setUser(localUser);
+        } else {
+          setShowLocalLogins(true);
+        }
+      }
+    }
+  }, [isIntegrationBranch]);
+
+  const isReadOnly = useCallback(() => {
+    return !user?.signInUserSession?.idToken?.payload?.["custom:cms_state"];
```
benjaminpaige, line 71: should we take the result of this and set it into a state variable isReadOnly in the checkAuthState function instead? or maybe doesn't need to be a function at all?

services/ui-src/src/hooks/authHooks/userProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+      } else {
+        const localUser = getLocalUserInfo();
+        if (localUser) {
+          setUser(localUser);
+        } else {
+          setShowLocalLogins(true);
+        }
+      }
+    }
+  }, [isIntegrationBranch]);
```
benjaminpaige, line 68: should we add Auth to this dependency array?

services/ui-src/src/hooks/authHooks/userProvider.tsx
```diff
@@ -0,0 +1,109 @@
...
+    checkAuthState();
+  }, [location, checkAuthState]);
+
+  const values: UserContextInterface = useMemo(
+    () => ({
+      user,
+      logout,
+      showLocalLogins,
+      loginWithIDM: authenticateWithIDM,
+      readOnly: isReadOnly(),
```
benjaminpaige, line 103: im worried 'readOnly' may get lost in the context, how do you feel about isReadOnlyUser?

services/ui-src/src/measures/2021/FUAAD/index.tsx
```diff
@@ -4,56 +4,72 @@ import * as Q from "./questions";
...
+        <CUI.Box
+          top={0}
+          left={0}
+          position={"fixed"}
+          width={"100vw"}
+          height={"100vh"}
+          zIndex={999}
+          // backgroundColor={"red"}
+          backgroundImage={"none"}
+        />
```
benjaminpaige, line 28: oh whoa cool, this is an interesting approach, is there suppose to be a semi-transparent background color? id love to see this in action. Also, If this is an approach we do want to go with we could add it into the MeasureWrapper so we only have to add it once.

services/ui-src/src/components/Layouts/StateLayout/index.tsx
benjaminpaige: i see why you changed this after reviewing below. how about '10' and change the one below to '9'?

services/ui-src/src/measures/2021/FUAAD/index.tsx
abelzer212: The issue with that is that it needs to block everything but not the header and not the footer, so the user can still use the navigations.

services/ui-src/src/components/Layouts/StateLayout/index.tsx
abelzer212: The Z-index is kind of arbitrary, we're just trying to make sure that the clear container is always at the very front of the page, by using 999-1000 we're making sure that any shenanigans that outside packages do to their component CSS will never prioritize what we have. We didn't want to do trial and error on what was the lowest high we could reach

services/ui-src/src/hooks/authHooks/userProvider.tsx
abelzer212: We had to use the useCallback in this circumstance because we were having issues with re-rendering, and a little more jankness. After a bit of research the internet said this is the way we have to do it.

services/ui-src/src/hooks/authHooks/userProvider.tsx
abelzer212: I feel like readOnly is in the context of the User information, and it's coming from the user provider, so I don't think it's confusing

services/ui-src/src/hooks/authHooks/userProvider.tsx
abelzer212: We're using Auth to check the current user, what do you mean by that?

services/ui-src/src/measures/2021/FUAAD/index.tsx
abelzer212: I moved it to measure wrapper, we didn't put in any color difference, presumably anyone who has admin access will know they aren't allowed to do anything, do you think we should add in like an opaque gray?

abelzer212: I made it opaque gray, I think that should be fine. We could also add a sticky note that says "You're not allowed to edit this page" or something like that if anyone gives us feedback that the gray isn't clear enough

services/ui-src/src/hooks/authHooks/userProvider.tsx
13bfrancis: I second what Ben said. How bout something like isAdmin, since this is what the check does in essence. readOnly doesn't make sense because it is not related to the context of the user. readOnly is a behavior that you are implementing, but is not information about the user

services/ui-src/src/hooks/authHooks/userProvider.tsx
daniel-belcher: As additional on this, we have another course, which was to drop the function down to the useMemo call, but if we were to keep it extracted from that we had to make it a callback or it would continuously re-render.

daniel-belcher: I believe the state variable would be an alternative as well, but as it only changes when the user object changes, the useMemo already changes, so just doing assertion there would be the same functionality without an additional hook

13bfrancis: I'm completely fine with this either way, but I will say from a readability sense that using useState is more readable

services/ui-src/src/hooks/authHooks/userProvider.tsx
daniel-belcher: As the check is currently checking for state information, "isStateUser" would be closer to contextual naming, and if a majority is for this, it should be a simple change

13bfrancis: Yeah, sorry. I got it flipped but yeah, something that is directly asserting true or false on what it is checking would be ideal

abelzer212: If it's a quick change I think it's fine to go to isStateUser but I don't think it's important enough to have an in depth argument about it.

benjaminpaige: i think isStateUser is the way to go 👍

services/ui-src/src/measures/2021/FUAAD/index.tsx
benjaminpaige: cool, yeah let's try it out. : )

services/ui-src/src/hooks/authHooks/userProvider.tsx
benjaminpaige: Also just realized this is the state attribute for state users, not the role. we should have both, a check for if the user is a state user, something like: user?.signInUserSession?.idToken?.payload?.["custom:cms_roles"] === "mdctqmr-state-user" 

and also one for the state (region): user?.signInUserSession?.idToken?.payload?.["custom:cms_state"]

benjaminpaige: i hate that these attributes are nested like this, but its what we've got to work with.

daniel-belcher: the state user check is a good point, we can definitely add that to the check

services/ui-src/src/components/MeasureButtons/index.test.tsx
```diff
@@ -48,18 +48,18 @@ describe("Test MeasureButtons", () => {
...
+  //     />
+  //   );
 
-    fireEvent.click(getByText(/Complete Measure/i));
+  //   fireEvent.click(getByText(/Complete Measure/i));
 
-    expect(mockChangeFn).toHaveBeenCalled();
-  });
+  //   expect(mockChangeFn).toHaveBeenCalled();
+  // });
```
benjaminpaige: either skip this or fix it or get rid of it

services/ui-src/src/components/MeasureButtons/index.tsx
```diff
@@ -13,6 +13,8 @@ export const MeasureButtons = ({
   handleSubmit,
   lastSavedText,
 }: Props) => {
+  const { readOnly } = useUser();
```
benjaminpaige, line 16: isStateUser

services/ui-src/src/hooks/authHooks/userProvider.tsx
benjaminpaige, line 103: lets change this to isStateUser,
also can we add the state attribute into this useUser as well

services/ui-src/src/measures/2021/FUAAD/index.tsx
```diff
@@ -4,56 +4,72 @@ import * as Q from "./questions";
...
-        </>
+      {readOnly && (
+        <CUI.Box
+          top={0}
+          left={0}
+          position={"fixed"}
+          width={"100vw"}
+          height={"100vh"}
+          zIndex={999}
+          // backgroundColor={"red"}
```
benjaminpaige: remove this line

services/ui-src/src/measures/2021/FUAAD/index.tsx
```diff
@@ -4,56 +4,72 @@ import * as Q from "./questions";
...
+          backgroundImage={"none"}
```
benjaminpaige: i thought you had a background color of gray.something here in the stuff i saw, did you get rid of it?

benjaminpaige (changes requested):
really if you can just add the state attribute to the useUser hook and change readOnly to isStateUser this would be good to go.

services/ui-src/src/measures/2021/FUAAD/index.tsx
abelzer212: Yeah it's on another branch

benjaminpaige (dismissed):

daniel-belcher (approved):

jdinh8124 (approved):

PR #353, opened 2022-01-03
fixed back button when inside a measure

jdinh8124:
## Purpose
Fixed the back button when inside a specific measure and should take you back to the list of measures

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -55,6 +55,10 @@ export const MeasureWrapper = ({ measure, name, year, measureId }: Props) => {
         breadcrumbItems={[
           { path: `/${params.state}/${year}`, name: `FFY ${year}` },
           {
+            path: `/${params.state}/${year}/${params.coreSetId}`,
+            name: "",
+          },
+          {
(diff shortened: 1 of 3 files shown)


services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -54,6 +54,10 @@ export const MeasureWrapper = ({ measure, name, year, measureId }: Props) => {
       <QMR.StateLayout
         breadcrumbItems={[
           { path: `/${params.state}/${year}`, name: `FFY ${year}` },
+          {
+            path: `/${params.state}/${year}/${params.coreSetId}`,
+            name: "",
+          },
```
benjaminpaige, line 60: Do we need this new object or can we just change the path of the first item in the array to /${params.state}/${year}/${params.coreSetId}

benjaminpaige: oh wait, i see why you did that, very clever. Maybe just add a comment or something above it so we know its just applicable to the back button.

jdinh8124: Sounds good to me

abelzer212 (approved):

PR #377, opened 2022-01-10
Api security

abelzer212:
#### Pull Request Creator Checklist

- [x] This PR has an associated issue or issues.
- [x] This PR meets all acceptance criteria for those issues.
- [x] This PR and linked issue(s) are adequately documented
- [x] This PR and linked issues(s) are a complete description of the changeset; an individual or team should be able to understand the issue(s) and changes by reading through this PR and it's links, with no further interaction.
- [x] Someone has been assigned this PR.
- [x] At least one person has been marked as reviewer on this PR.


services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
+import { APIGatewayProxyEvent } from "aws-lambda";
+
+const authErrorHandler = (
+  state: String,
+  userState: String,
+  userRole: String,
+  operationType: String
```
benjaminpaige, line 7: why is string capitalized?

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (operationType === "GET" || operationType === "LIST") {
+    if (
+      userRole.includes("state") &&
+      state.toLowerCase() !== userState.toLowerCase()
+    ) {
+      return 403;
+    }
+  }
+  return 200;
+};
```
benjaminpaige, line 33: would it make more sense to switch on the operationType, also would 'method' be more appropriate than 'operationType'?

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const errorHandler = (
+  event: APIGatewayProxyEvent,
+  operationType: String
+) => {
+  if (!event.pathParameters) return 400;
```
benjaminpaige, line 39: can we just add this to the if block on ln 40

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+      userRole.includes("state") &&
+      state.toLowerCase() !== userState.toLowerCase()
+    ) {
+      return 403;
+    }
+  }
+  return 200;
+};
+
+export const errorHandler = (
```
benjaminpaige, line 35: instead of 'errorHandler', how do you feel about eventValidator or requestValidator or something along those lines?

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+    return 400;
+
+  if (operationType !== "LIST" && !event.pathParameters.coreSet) return 400;
+
+  return authErrorHandler(
+    event.pathParameters.state,
+    // @ts-ignore
+    event.headers.user_state,
+    event.headers.user_role,
+    operationType
```
benjaminpaige, line 54: can we use a single object instead of four arguments here?

services/app-api/handlers/authUtils/checkAuth.ts
benjaminpaige, line 7: should we just use a single parameter object with a single interface?

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+  );
+};
+
+export const measureErrorHandler = (
```
benjaminpaige, line 58: these arent really handling the errors, moreso just validating the events/requests. Is there a better naming pattern for these.

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+  event: APIGatewayProxyEvent,
+  operationType: String
+) => {
+  if (!event.pathParameters) return 400;
```
benjaminpaige, line 62: this can go in the if block below

services/app-api/handlers/coreSets/create.ts
```diff
@@ -3,22 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
 
 export const createCoreSet = handler(async (event, context) => {
-  if (!event.pathParameters) return; // throw error message
-  if (
-    !event.pathParameters.state ||
-    !event.pathParameters.year ||
-    !event.pathParameters.coreSet
-  )
-    return; // throw error message
+  const errorCode = errorHandler(event, "POST");
```
benjaminpaige, line 10: it seems like all we are really doing here is checking to see if the request is valid. maybe this could just return a true or false? and if it return false return a generic 400 error.

services/app-api/handlers/coreSets/create.ts
```diff
@@ -3,22 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
+  if (errorCode !== 200) {
+    return {
+      statusCode: errorCode,
+      body: JSON.stringify({
+        error: "Failure: HTTP Status Code ",
```
benjaminpaige, line 15: at some point we should make these errors a little more verbose, until then we really only need the failed request and statusCode.

services/app-api/handlers/coreSets/create.ts
```diff
@@ -3,22 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
 
   // The State Year and ID are all part of the path
-  const state = event.pathParameters.state;
-  const year = event.pathParameters.year;
-  const coreSet = event.pathParameters.coreSet;
-  const type = coreSet?.substring(0, 1);
+  const state = event!.pathParameters!.state!;
+  const year = event!.pathParameters!.year!;
+  const coreSet = event!.pathParameters!.coreSet!;
+  const type = coreSet!.substring(0, 2);
```
benjaminpaige, line 25: either this needs to stay at substring(0, 1) or we need to find another way to search for measures to add based on coreset abbr.

services/app-api/handlers/coreSets/create.ts
```diff
@@ -3,22 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
+  }
 
   // The State Year and ID are all part of the path
-  const state = event.pathParameters.state;
-  const year = event.pathParameters.year;
-  const coreSet = event.pathParameters.coreSet;
-  const type = coreSet?.substring(0, 1);
+  const state = event!.pathParameters!.state!;
+  const year = event!.pathParameters!.year!;
+  const coreSet = event!.pathParameters!.coreSet!;
```
benjaminpaige, line 24: can we cast a type onto the event or use the APIGatewayProxyEvent. we really shouldn't be using non-null assertion operators like this.

services/app-api/handlers/coreSets/create.ts
```diff
@@ -3,22 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
+  const type = coreSet!.substring(0, 2);
+
   const coreSetQuery = await getCoreSet(event, context);
   const coreSetExists = !!Object.keys(JSON.parse(coreSetQuery.body)).length;
```
benjaminpaige, line 28: ideally this would be a try catch, but if we arent going to throw an error in getCoreSet is there a better way of checking if a coreset exists?

services/ui-src/src/hooks/authHooks/useApi.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ApiContext } from "./ApiProvider";
+import { API } from "aws-amplify";
+import { useEffect } from "react";
+import config from "config";
+import { useUser } from "hooks/authHooks";
+
+export const useApi = () => {
+  const context = useContext(ApiContext);
+
+  const userInfo = useUser();
```
benjaminpaige, line 11: can we add a User type into userInfo?

services/ui-src/src/hooks/authHooks/useApi.tsx
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(() => {
+    API.configure({
+      endpoints: [
+        {
+          name: "coreSet",
```
benjaminpaige, line 16: do we need to add the 'measures' endpoint to this array as well?

services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (!state || !userState || !userRole || !operationType) {
+    return 403;
+  }
+  if (
+    operationType === "POST" ||
+    operationType === "PUT" ||
+    operationType === "DELETE"
+  ) {
+    if (
+      !userRole.includes("state") ||
```
benjaminpaige, line 18: we should be more specific here, maybe even use a user role enum

services/ui-src/src/components/Footer/index.tsx
```diff
@@ -4,6 +4,7 @@ import * as Libs from "libs";
 export function Footer(): JSX.Element {
   return (
     <CUI.Box
+      zIndex={1001}
```
benjaminpaige, line 7: zIndex={3} should be fine

services/ui-src/src/hooks/authHooks/ApiProvider.tsx
```diff
@@ -0,0 +1,38 @@
...
+              // @ts-ignore
+              user_state: userInfo!.userState!,
+              // @ts-ignore
+              user_role: userInfo!.user!.role!,
+            };
+          },
+        },
+      ],
+    });
+  }, []);
```
benjaminpaige, line 33: prob don't need this in both useApi and ApiProvider right?

benjaminpaige (changes requested):
among other things, we seem to be doing something like this: 
`  const errorCode = something(event, method);
  if (errorCode !== 200) {
    return {
      statusCode: errorCode,
      body: JSON.stringify({
        error: "Failure: HTTP Status Code ",
        errorCode,
      }),
    };
  }`
  
  in a lot of places, i wonder if we can move it up a level so we only need to do it once. If we can do that we can then assert types to these handlers so we know exactly what the event looks like and don't have to validate properties all over the place.

services/app-api/handlers/authUtils/checkAuth.ts
jdinh8124: I'm assuming the roles are the same as the front end, heres a snippet of the enum:

`
export enum UserRoles {
STATE = "mdctqmr-state-user",
ADMIN = "mdctqmr-approver",
}
`

services/ui-src/src/hooks/authHooks/ApiProvider.tsx
jdinh8124, line 33: src/hooks/authHooks/ApiProvider.tsx
  Line 35:6:  React Hook useEffect has missing dependencies: 'StateInfo' and 'userInfo'. Either include them or remove the dependency array  react-hooks/exhaustive-deps

services/ui-src/src/hooks/authHooks/useApi.tsx
```diff
@@ -0,0 +1,36 @@
...
+              //@ts-ignore
+              user_state: userInfo!.userState!,
+              //@ts-ignore
+              user_role: userInfo!.user!.role!,
+            };
+          },
+        },
+      ],
+    });
+  }, []);
```
jdinh8124, line 30: Another pipeline failure due to this line
src/hooks/authHooks/useApi.tsx
  Line 30:6:  React Hook useEffect has missing dependencies: 'userInfo.user.role' and 'userInfo.userState'. Either include them or remove the dependency array  react-hooks/exhaustive-deps

13bfrancis:
I think there is a wrapper function for all the lambdas. Maybe in there?

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -15,6 +15,7 @@ interface Data {
   year: string;
   coreSet: CoreSetAbbr;
 }
+//
```
jdinh8124, line 18: Unused comment

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: I couldn't quite get the switch to work so I just did the if statements

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: I'm not attached to names, seems like an easy change

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: addressed

services/ui-src/src/hooks/authHooks/ApiProvider.tsx
abelzer212: fixed

services/ui-src/src/hooks/authHooks/useApi.tsx
abelzer212: fixed

services/ui-src/src/hooks/authHooks/ApiProvider.tsx
abelzer212: Yeah, I was getting it working, I'll take that out now and see if it still works correctly

services/ui-src/src/views/StateHome/index.tsx
abelzer212: fixed

services/app-api/handlers/coreSets/create.ts
abelzer212: fixed it

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: Yup! done

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: changed to validator

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: done

services/app-api/handlers/coreSets/create.ts
abelzer212: Yeah, I was debating that, but I like to give specific error codes if I can for debugging purposes

services/app-api/handlers/coreSets/create.ts
abelzer212: Yeah, you and I talked about doing an error layer above every handler call, but I'll break that out into its own story

services/app-api/handlers/coreSets/create.ts
abelzer212: I was going to just ts-ignore all of these because we are handling the case where the sub-objects don't exist, but I'd rather not ts-ignore if I don't have to. I think the assertions are ok because we're explicitly checking existence in the validation handler

services/app-api/handlers/coreSets/create.ts
abelzer212: We're throwing an error in the lines below this. I don't know how we would check if it exists without querying to see if it exists.

services/ui-src/src/hooks/authHooks/useApi.tsx
abelzer212: For functionality? No, everything is currently passing through the coreset endpoint. For clarity, yeah that would probably be good to separate the two calls.

services/app-api/handlers/authUtils/checkAuth.ts
abelzer212: I'll snag that and implement it over here.

services/ui-src/src/components/Footer/index.tsx
abelzer212: whoops that was a weird branching issue

services/ui-src/src/hooks/authHooks/ApiProvider.tsx
abelzer212: and it's gone

services/ui-src/src/hooks/authHooks/useApi.tsx
abelzer212: I ripped this out so we don't need it anymore

benjaminpaige (approved):

PR #394, opened 2022-01-12
Adult Core Set Question Qualifiers

benjaminpaige:
In addition to the ui build out for this page there are some additional features here: 

Connected qualifiers questions to db
User can submit data
Will retrieve and load previously saved data.
Shows last submitted time if previously submitted (relative to current time) i.e. “Submitted less than a minute ago” in sub header.
Shows “submitted” and checkmark on measures list page if qualifiers have been submitted (“incomplete” otherwise)
Disabled submit button if qualifiers have not been touched since last submit event
Note: there are currently no limits to data validation when completing core qualifiers. submitting some or all data is allowed.


services/ui-src/src/types.ts
```diff
@@ -10,3 +10,14 @@ export enum UserRoles {
   STATE = "mdctqmr-state-user",
   ADMIN = "mdctqmr-approver",
 }
+
+export enum MeasureStatus {
+  COMPLETE = "complete",
+  INCOMPLETE = "incomplete",
+}
```
jdinh8124, line 17: Does the "NA" status the same as the complete status?

jdinh8124 (dismissed):
Lots of files changes, but everything looks good to me and nothing stands out as an issue. Thank you for cleaning up a lot of the files.

services/ui-src/src/types.ts
benjaminpaige: I think that's the ffy reporting status, but good call out, ill find a place for that.

abelzer212 (approved):
I ended up touching all of the core set api hooks so I think this will conflict with some of that, and for the measures you will need to change some of the arguments to get this to work, but I think that Daniels branch will be doing that already so it should be fine

abelzer212 (dismissed):

abelzer212 (approved):

abelzer212 (approved):

abelzer212 (approved):

jdinh8124 (approved):

PR #405, opened 2022-01-15
Api rewrite

abelzer212:
## Purpose

Dev is failing right now because CORS on the master branch doesn't allow custom headers, I've removed the custom headers and put regular CORS rules back in place. The User State and User Role are now passed in in the body of the request

### DIFF
--- a/services/app-api/handlers/authUtils/checkAuth.ts
+++ b/services/app-api/handlers/authUtils/checkAuth.ts
@@ -39,13 +39,16 @@ const authErrorHandler = (
 
 export const eventValidator = (
   event: APIGatewayProxyEvent,
-  operationType: String
+  operationType: string
 ) => {
+  if (!event.body) return 400;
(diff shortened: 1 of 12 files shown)


services/app-api/handlers/authUtils/checkAuth.ts
```diff
@@ -39,46 +39,50 @@ const authErrorHandler = (
 
 export const eventValidator = (
   event: APIGatewayProxyEvent,
-  operationType: String
+  operationType: string
 ) => {
+  if (!event.body) return 400;
+  const { userRole, userState } = JSON.parse(event.body);
   if (
+    !event.body ||
```
daniel-belcher, line 47: Is this event.body check necessary if it would return right above? Absolutely not a blocker.

daniel-belcher (approved):
Rough that our GET requests have to change to POST for user validation, but otherwise this all looks good.

PR #422, opened 2022-01-20
Auth fix 4

benjaminpaige:
replacement for the last auth-fix branch pr

resolved the cors issue with custom headers
updates to correct rest methods
uses token to validate user in api
removed local login buttons


13bfrancis (approved):
Looks good

PR #468, opened 2022-01-31
change baseUrl for local and CI automated testing

benjaminpaige:
remove hardcoded baseUrl

use localhost for local 

use application endpoint to run tests in ci


aalousi (commented):
This should work i believe, looks great. Thanks Ben. My only question is as follows: for the cypress.json. if we point towards a specific env. is it going to be overridden in CI to what ever env that we are in?

13bfrancis (dismissed):

benjaminpaige:
> 

Good question @aalousi yes it will be overriden in the CI to whatever the deployed url is for that branch/environment. So it should always run based on the current state of the code. This will make it easy to add id's or change how we are selecting elements. But lets make sure that everyone on the QE team can get the application up and running locally on their machine today without any issues.

PR #475, opened 2022-01-31
Ietad

abelzer212:
IET Measure


services/ui-src/src/measures/2021/IETAD/validation/types.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    GeographyOptions: string[];
+    UrbanGeographyRate: AggregateRate;
+    RuralGeographyRate: AggregateRate;
+    AddtnlGeographyDesc: string;
+    AddtnlGeographyRate: AggregateRate;
+
+    ACAGroupRate: AggregateRate;
+  }
+}
```
benjaminpaige, line 311: Can double check to make sure that all of the types in this file are used within the measure? Obviously we want to make sure these are all represented in the measure when we hand this to mathmatica

abelzer212: Yes this is part of the OPM

jdinh8124 (approved):
Looks good and the global validations have already helped me in tons of other measures

services/ui-src/src/measures/2021/IETAD/validation/types.ts
```diff
@@ -0,0 +1,311 @@
...
+    "DataSource-Administrative-Other"?: string;
+    "DataSource-Administrative-Other-Explanation"?: string;
+    "DataSource-Electronic": string;
+    "DataSource-Electronic-Explanation": string;
+    "DataSource-Other": string;
+    "DataSource-Other-Explanation": string;
+    "DataSource-Hybrid"?: string[];
+    "DataSource-Hybrid-Other"?: string;
+    "DataSource-Hybrid-Other-Explanation"?: string;
+    "DataSource-Hybrid-MedicalRecord-DataSoruce"?: string;
```
jdinh8124, line 47: I think some of the hybrid data sources could be removed

jdinh8124 (dismissed):

PR #515, opened 2022-02-10
Oy2 9954 e2e

benjaminpaige:
fixed rate calc error
add common cypress commands
add testing for measure
add id to checkbox
etc


tests/cypress/cypress/integration/oy2-9954_PQI01AD.spec.js
```diff
@@ -0,0 +1,278 @@
...
+  before(() => {
+    cy.visit("/");
+    cy.login();
+    cy.goToAdultMeasures();
+    cy.goToMeasure("PQI01-AD");
+  });
+
+  it("Ensure correct sections display if user is/not reporting", () => {
+    cy.displaysSectionsWhenUserNotReporting();
+    cy.displaysSectionsWhenUserIsReporting();
```
benjaminpaige, line 14: i created these common functions so we can reuse them in other places. they live in support/commands

aalousi (approved):
Looks great!

PR #526, opened 2022-02-12
Update clear data functionality

benjaminpaige:
the methods.reset() method doesnt reset the checkboxes or radio buttons. Without actually rewriting those components this may be a good option, it submits an empty data object and then goes back a page. 🤷 

also added that callback to the submit functionality because i think that makes sense

and updated a cypress command to clear measures and then go back into the measure.

Anyways, if you wanted to go this direction id be good with it. Just a suggestion.

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -7,6 +7,8 @@ import { useParams } from "react-router-dom";
 import { useGetMeasure, useUpdateMeasure } from "hooks/api";
 import { AutoCompletedMeasures, CoreSetAbbr, MeasureStatus } from "types";
 import { v4 as uuidv4 } from "uuid";
+import { useNavigate } from "react-router-dom";
+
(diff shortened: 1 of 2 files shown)


charmcitygavin (approved):

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -91,7 +94,13 @@ export const MeasureWrapper = ({
   };
 
   const handleClear = () => {
-    submitDataToServer({ data: {} }, MeasureStatus.INCOMPLETE);
+    submitDataToServer({
+      data: {},
+      status: MeasureStatus.INCOMPLETE,
+      callback: () => {
+        navigate(-1);
+      },
```
jphechter, line 102: Wasn't familiar with this, but using it in both `handleClear()` and `handleSubmit()` makes a lot of sense.

jphechter (approved):
Noice! Thank you! 👍

PR #527, opened 2022-02-12
Parallel e2e

benjaminpaige:
adding parallelization to cypress tests in ci
currently looking into how to achieve this without naming files in test dir specifically.


aalousi (commented):
This looks amazingly amazing. with that, we can create smoke tests, separated in dif. containers. and a regression. Comment out, add new tests as we go. and create containers for separate components or a full regression as well. Lets circle back for a quick meeting on Monday to take a deeper dive into it! but looks great! Thanks Ben.

aalousi (approved):

gitsierra (approved):

PR #643, opened 2022-03-09
Add e2e tests for ammad

benjaminpaige:


aalousi (commented):
looks good, please add it to the deploy.yml as well!

aalousi (approved):

PR #665, opened 2022-03-11
fix date

abelzer212:
remove helper text from month

### DIFF
--- a/services/ui-src/src/components/MonthPicker/index.test.tsx
+++ b/services/ui-src/src/components/MonthPicker/index.test.tsx
@@ -160,8 +160,8 @@ describe("Testing MonthPicker", () => {
     );
 
     // change month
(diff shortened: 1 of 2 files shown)


jphechter (dismissed):
👍

services/ui-src/src/components/MonthPicker/index.test.tsx
```diff
@@ -88,7 +88,7 @@ describe("Testing MonthPicker", () => {
     );
     expect(getByDisplayValue("11"));
     expect(getByDisplayValue("2011"));
-    expect(getByText(/Month \(mm\):/i));
+    expect(getByText(/Month:/i));
```
jphechter, line 91: Good catch! Saw this in the deploy fail, was about to point to it.

jphechter (approved):
_**NOW**_ its a 👍

PR #702, opened 2022-03-18
Fix rounding error

benjaminpaige:


abelzer212 (approved):

benjaminpaige:
@abelzer212 @13bfrancis just want to double check this wont break anything? wdyt?

abelzer212 (approved):
I banged on it for a few different measures for OPM, OMS, and PM, and it behaved as expected

PR #1302, opened 2022-06-14
Oy2 17993 automation

jdinh8124:
## Purpose
Added basic automation for the OEVCH measure


benjaminpaige:
this is a 2022 measure. closing as we are going a diff direction for testing these

PR #1322, opened 2022-06-21
Upgrage chakra and react

benjaminpaige:
update chakra-ui and react and resolve errors


13bfrancis (dismissed):
This is so exciting!!

PR #1325, opened 2022-06-22
Same PR but in a different place

abelzer212:
Move the tweek to the new branch


benjaminpaige (approved):
And a rate formula test 🤯

PR #1505, opened 2022-09-02
add coreset to completed check

gmrabian:
## Purpose

Fixes an issue where people couldn't submit when they completed all measures _and_ the coreset (because then the values were unequal)

According to David all measure sets should have a coreset so this check safely works for those, as well as when someone adds measures (because # completed must match # of measures + 1)


braxex (approved):
Sounds reasonable to me.

juliuscaesar (approved):

PR #1535, opened 2022-09-29
MDCT-1961: Fetching on Focus Fix

karla-vm:
## Description

@bodnarbm found an issue with how `react-query` re-fetches a query on window focus, which was causing a refresh of "stale" data in the QMR measure forms. Now, we're explicitly skipping that re-fetch.

### How to test

./dev local
- Enter any measure
- Start typing data
- Navigate to a different tab

The data should still be there 🎉 

### Changed Dependencies

N/A

## Code author checklist

- [ ] I have added [thorough](https://bit.ly/3zPrxuZ) tests
- [ ] I have added analytics, if necessary
- [ ] I have updated the documentation, if necessary

## Reviewer checklist (two different people)

- [ ] I have done the deep review and verified the items in the above checklist are g2g
- [ ] I have done the lgtm review

### DIFF
--- a/services/ui-src/src/measures/2021/CommonQuestions/DataSource/__snapshots__/index.test.tsx.snap
+++ b/services/ui-src/src/measures/2021/CommonQuestions/DataSource/__snapshots__/index.test.tsx.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Test the global DataSource component (Custom Structure) Component renders with correct content 1`] = `
-<body>
+<div>
   <div
     class="css-pb69ky"
   >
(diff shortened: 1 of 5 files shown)


services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -205,19 +205,18 @@ export const MeasureWrapper = ({
...
-              (params.coreSetId?.split("_")?.[0] ??
-                params.coreSetId) as CoreSetAbbr
+              (coreSetId?.split("_")?.[0] ?? coreSetId) as CoreSetAbbr
             ]?.formData
           : undefined
       );
     }
     // default loaded data reset
-    else {
+    else if (!methods.formState.isDirty) {
```
gmrabian, line 216: with this change does this part work as an `else` now? Because you still have the original issue to fix, no? Or is this what was causing it?

karla-vm: Good call, I tested it without the explicit condition and it works as expected !

gmrabian (approved):

juliuscaesar (approved):

PR #1539, opened 2022-09-30
VAL Release (30 September 2022)

karla-vm:
`master` -> `val`


gmrabian (commented):
Looks good. Do we want to try to fix the cypress tests that failed in master before promoting?

karla-vm:
> Looks good. Do we want to try to fix the cypress tests that failed in master before promoting?

Good call, now they're cleaned up! ✅

gmrabian (approved):

PR #1576, opened 2022-12-29
Val

TheDanMiller:
## Description

Dynamo power up moving from val --> prod. Run in val looks like 

`REPORT RequestId: b89ef476-6ba3-4506-9708-de28c7fff248	Duration: 2014.99 ms	Billed Duration: 2015 ms	Memory Size: 4096 MB	Max Memory Used: 167 MB	Init Duration: 467.37 ms	
`

Can see the increased memory limit. Definitenly do not need that much memory in prod, this is just a safeguard though to buy us time until we can make this right.


braxex:
Approved from mobile?

braxex (approved):
ah this is how. cool.

karla-vm (approved):

BearHanded (approved):
✨ _Season's Greetings!_ ✨

PR #1665, opened 2023-03-08
fix the text for other performance measures

britt-mo:
## Summary
This is the same fix as before, to get a line break for the formula. This was overlooked in my last PR.
### How to test
Go to AAB-AD and get the other performance measures to show up. You'll see that the string has a line break and not a literal `<br>` tag in the browser.

> 
>  "Enter a number for the numerator and the denominator. The measure is reported as an inverted rate. 
> <br/> The formula for the Rate = (1 - (Numerator/Denominator)) x 100"

Test the same thing inside the Optional Measure Strat section as well 

### Author checklist

- [ ] I have updated the documentation, if necessary


BearHanded (approved):

BearHanded (approved):

karla-vm (approved):
LGTM!

PR #1682, opened 2023-04-03
[MDCT–2340] Complete Measure when Invalid - Save correct data

BearHanded:
### Description
Fix the Measure Wrapper "Complete Measure" functionality to submit the most recent data, rather than all preset data chunks.

When an invalid measure is manually set to "Complete" by clicking through the modal, a bad state could be sent alongside the request. Moved it to allow the form itself to handle the submission after validation. See testing instructions for more details.

### Related ticket(s)

MDCT-2340

---
### How to test
The previous bad state could be triggered as followed:
* Pick a measure with nested items. The child measure WCC-CH, for example.
* Begin populating the measure, but leave some part, like the date fields incomplete to trigger validation warnings (important for the next part)
* In the Data sources, fill some root and nested fields
* Click Save. The database should reflect a populated DataSource, and DataSourceSelections.
* Uncheck the root levels, complete measure, and when the modal pops up, select yes.
* All the DataSource and DataSourceSelections will be cleared!

Before this update, the DataSourceSelections would remain populated.
The original request also found this happening when a measure is set not to reporting, nested options for reasons for not reporting could get out of sync.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -354,9 +354,9 @@ export const MeasureWrapper = ({
 
   const handleValidationModalResponse = (continueWithErrors: boolean) => {
     setShowModal(false);
+    if (!continueWithErrors) return;
 
-    if (continueWithErrors) {
-      const data = methods.getValues();
+    const manualSubmit = (data: any) => {
       submitDataToServer({
         data,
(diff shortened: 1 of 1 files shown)


karla-vm (approved):

braxex (approved):
Haven't run it but code looks good.

PR #1683, opened 2023-04-03
fix radio button bug

britt-mo:
### Description
In `Measure Specifications` the 'Other' radio button was not functioning as expected when clicked. This was due to a `useEffect` that was messing up the life cycle. 

### Related ticket(s)

MDCT-2338

---
### How to test
Go to any of the measures in 2023, go down to `Measure Specifications` and select the radio button with the text `Other` (don't select the radio button above it. See that it expands to show a text box and an area to upload documents.

### Important updates

---
### Author checklist

---


BearHanded (approved):
Tested this change locally, but I would appreciate another pair of eyes just to make sure we're not missing anything here.

karla-vm (approved):

PR #1725, opened 2023-04-13
Update HEDIS options for 2023

benmartin-coforma:
### Description
Add an option for HEDIS MY 2023, and remove the option for HEDIS 2020.

Since all HEDIS options now use the MY naming scheme, this PR also removes the clarifying text.

### Related ticket(s)
MDCT-2358

I believe this will also address MDCT-2368; 2358 is for adult measures and 2368 is for child measures, but as far as I can tell they use the same code.

---
### How to test
1. Log into QMR
2. Choose the 2023 reporting year
3. Open the adult core sets
4. Open any measure
5. For the third question ("Measurement Specification"), ensure the first radio button is checked
6. Note the absence of any explanatory text regarding the naming conventions for HEDIS 2020 vs HEDIS MY 2020.
7. Expand the nested dropdown; note that the options are correct.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/measures/2023/shared/CommonQuestions/MeasurementSpecification/index.tsx
+++ b/services/ui-src/src/measures/2023/shared/CommonQuestions/MeasurementSpecification/index.tsx
@@ -9,11 +9,6 @@ const HEDISChildren = () => {
 
   return (
     <>
-      <CUI.Text key="measureSpecDescriptor" size="sm" pb="3">
-        NCQA, the measure steward, changed its naming convention. HEDIS MY 2020
-        refers to a different federal fiscal year (FFY) than HEDIS 2020. Please
-        note the FFY Core Set specification below.
-      </CUI.Text>
       <QMR.Select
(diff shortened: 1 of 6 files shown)


services/ui-src/src/measures/2023/shared/CommonQuestions/MeasurementSpecification/index.tsx
```diff
@@ -9,16 +9,15 @@ const HEDISChildren = () => {
```
ailZhou, line 9: Reviews and tested in browser, looks good but did get a slight warning from compiler about the import CUI not being used. Minor request to remove "import * as CUI from "@chakra-ui/react";" from this file to appease the complier

BearHanded: Yea, looks like this is stopping it from deploying as well

ailZhou (approved):

BearHanded (approved):

PR #1764, opened 2023-04-26
Enable ids 2023

ailZhou:
### Description
Large update to fix issue in QMR where the label information is being utilized as if it's an id. Any updates made to the label will cause the database identification to change, which creates an issue downstream. 

Only in year 2023 onward, no changes for the previous years.

### Related ticket(s)

MDCT-2268
MDCT-2429

---
### How to test

1) Run QMR and spin up local dynamodb
2) Open rateLabelText.tsx to use as a comparison when viewing the key names in the database.
3) Sign in using account stateuserWA@test.com (this contains AD, CH & HH measures).
4) Select Year 2023
5) This update applies to all measures, so select any one of them
6) For the first 3 questions, select the first radio button. This is will the enable the Performance Measure view
7) Scroll down to Performance Measure and fill out the fields, Numerator & Denominator (Rate will auto populate)
8) Scroll down to Optional Measure Stratification and fill out with any fake data in any of the categories.
9) Click Validate Measure
10) Key fields in local dynamodb for the measure you've just validated should show names that are the same as the IDs field in rateLabelText.tsx

### Important updates

N/A

---
### Author checklist

---


britt-mo (dismissed):
Tested in the browser and db, worked as expected.

services/ui-src/src/measures/2023/rateLabelText.ts
```diff
@@ -387,254 +385,254 @@ export const data = {
...
-        "categories": []
+        "categories": [{"id":"CHL-AD-CAT-A", "label": "", "text":""}]
     },
     "CHL-CH": {
         "qualifiers": [
             {
                 "label": "Ages 16 to 20",
                 "text": "Ages 16 to 20",
-                "id": "Ages16to20"
+                "id": "QUAL-B"
```
benmartin-coforma, line 463: CONVENTION: This ID would be `QUAL-A` if it followed the naming pattern. Per Aileen, this won't cause any functional problems, but it may be good to adhere to the convention anyway.

services/ui-src/src/measures/2023/shared/CommonQuestions/OptionalMeasureStrat/ndrSets.tsx
```diff
@@ -100,13 +104,13 @@ const TotalNDRSets = ({
   const { qualifiers, categories } = usePerformanceMeasureContext();
   const totalQual = qualifiers.slice(-1)[0];
 
-  if (categories.length) {
+  if (categories.length && categories?.filter((item) => item.label).length) {
```
benmartin-coforma, line 107: NIT: I think `.some` would be a little clearer here
```suggestion
  if (categories.length && categories.some((item) => item.label)) {
```

services/ui-src/src/measures/2023/shared/CommonQuestions/OptionalMeasureStrat/ndrSets.tsx
```diff
@@ -128,11 +132,12 @@ const TotalNDRSets = ({
   return (
     <CUI.Box>
       <CUI.Divider key={`totalNDRDivider`} mt={2} mb={5} />
-      {categories.length > 0 && (
-        <CUI.Heading size={"sm"} key={`totalNDRHeader`}>
-          {totalQual}
-        </CUI.Heading>
-      )}
+      {categories.length > 0 &&
+        categories?.filter((item) => item.label).length > 0 && (
```
benmartin-coforma, line 136: Same suggestion as above, here

```suggestion
        categories.some((item) => item.label) && (
```

services/ui-src/src/measures/2023/shared/CommonQuestions/OptionalMeasureStrat/ndrSets.tsx
```diff
@@ -328,7 +330,10 @@ const useAgeGroupsCheckboxes: CheckBoxBuilder = (name) => {
 
   const qualRates = useQualRateArray(name);
   const standardRates = useStandardRateArray(name);
-  const rateArrays = !categories.length ? qualRates : standardRates;
+  const rateArrays =
+    !categories.length || !categories?.filter((item) => item.label).length
```
benmartin-coforma, line 334:
```suggestion
    !categories.length || !categories.some((item) => item.label)
```

services/ui-src/src/measures/2023/shared/CommonQuestions/PerformanceMeasure/index.tsx
```diff
@@ -162,7 +162,10 @@ const QualifierNdrSets = ({
 const PerformanceMeasureNdrs = (props: NdrSetProps) => {
   let ndrSets;
 
-  if (props.categories?.length) {
+  if (
+    props.categories?.length &&
+    props.categories?.filter((item) => item.label).length
```
benmartin-coforma, line 167:
```suggestion
    props.categories.some((item) => item.label)
```

services/ui-src/src/measures/2023/shared/globalValidations/validateEqualCategoryDenominators/index.ts
```diff
@@ -20,14 +21,18 @@ const _validation: UVF = ({
...
     for (const [j, rate] of rateSet.entries()) {
       if (rate && rate.denominator) {
         denominatorArray.push(rate.denominator);
         locationArray.push(
-          !!categories?.length && categories[0] !== SINGLE_CATEGORY
-            ? categories![i]
-            : qualifiers![j]
+          !!categories?.length &&
+            categories[0].label !== SINGLE_CATEGORY &&
+            categories?.filter((item) => item.label).length
```
benmartin-coforma, line 33:
```suggestion
            categories.some((item) => item.label)
```

services/ui-src/src/measures/2023/PCRHH/questions/PerformanceMeasure.tsx
```diff
@@ -91,7 +97,11 @@ const QualifierNdrSets = ({
 const PerformanceMeasureNdrs = (props: NdrSetProps) => {
   let ndrSets;
 
-  if (props.categories?.length) {
+  //if there is a category and the category labels are filled out, create the NDR using the categories array
+  if (
+    props.categories?.length &&
+    props.categories?.filter((item) => item.label).length
```
benmartin-coforma, line 103:
```suggestion
    props.categories?.some((item) => item.label)
```

services/ui-src/src/measures/2023/shared/CommonQuestions/DeviationFromMeasureSpecification/index.tsx
```diff
@@ -137,20 +137,21 @@ export const DeviationFromMeasureSpec = ({
       const topLvlOptions: TopLevelOptions = [];
       const { rates } = watchPerformanceMeasure;
 
-      if (rates.singleCategory) {
-        // handle for PCR-XX measures
+      //by this point in the system, there should be at least 1 key in rates
+      //if there are no categories, the key singleCategory will be used instead
```
BearHanded, line 141: is this still true? I thought we did away with the singleCategory marker

services/ui-src/src/measures/2023/shared/CommonQuestions/PerformanceMeasure/index.test.tsx
```diff
@@ -124,7 +124,12 @@ describe("Test the PerformanceMeasure RateComponent prop", () => {
 
   test("(PCR-XX) Ensure component renders", () => {
     // modifying data to be easier to check
-    PCRData.qualifiers = PCRData.qualifiers!.map((qual) => `qual ${qual}`);
+    // PCRData.qualifiers = PCRData.qualifiers!.map((qual) => `qual ${qual}`);  //ORIGINAL
```
BearHanded, line 127: Do we need to save this original and changed comment?

BearHanded (changes requested):

services/ui-src/src/measures/2023/shared/CommonQuestions/DeviationFromMeasureSpecification/index.tsx
```diff
@@ -95,7 +94,8 @@ export const getLowLvlDeviationOptions = ({
   return qualifiers
     .sort((a: any, b: any) => b.label! - a.label!)
```
benmartin-coforma, line 95: COMMENT: I think this sort does nothing? Since the labels are strings, and a string minus another string is `NaN`. I believe we would need something like this:

```ts
.sort((a, b) => {
  if (a.label && b.label) {
    return a.label.localeCompare(b.label);
  }
  else if (a.label) {
    return -1;
  }
  else {
    return 1;
  }
})
```

This is non-blocking; I don't think this needs to be addressed before merging. I guess I'm just curious if it _does_ actually work.

services/ui-src/src/measures/2023/shared/globalValidations/validateEqualCategoryDenominators/index.ts
```diff
@@ -20,14 +21,18 @@ const _validation: UVF = ({
   const locationArray: string[] = [];
   const denominatorArray: string[] = [];
 
+  console.log(qualifiers?.length);
+
```
benmartin-coforma, line 25: This `console.log` is trying to sneak its way in 👀

```suggestion
```

services/ui-src/src/utils/testUtils/2023/validationHelpers.ts
```diff
@@ -0,0 +1,312 @@
...
+  denominator: "",
+  rate: "",
+  label: "Empty Test Label",
+};
+
+/**
+ * Helper function to prep oms validation test data  by slotting test data in qualifier order
+ *
+ * @param categories should always at least contain "singleCategory"
+ * @param qualifiers a non-negotiable string array
```
benmartin-coforma, line 85:
```suggestion
 * @param qualifiers a non-negotiable LabelData array
```

services/ui-src/src/measures/2023/shared/globalValidations/validatePartialRateCompletion/index.ts
```diff
@@ -128,7 +129,11 @@ export const validatePartialRateCompletionOMS =
               ...label,
             ])}`,
             rateData: rateData?.[singleValueFieldFlag],
-            categories: !!(isOPM || categories[0] === SINGLE_CATEGORY)
+            categories: !!(
+              isOPM ||
+              categories[0].label === SINGLE_CATEGORY ||
+              !categories[0].label
+            )
```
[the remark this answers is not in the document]
benmartin-coforma: Same question for lines 152-156 in this file.

benmartin-coforma (commented):
Overall, this is really good work.

I found two files which still seem to be treating qualifiers as strings:
1. services/ui-src/components/PCRRate/index.test.tsx
   * Aileen and I discussed this; since the tests are passing it is probably a nonissue
2. tests/cypress/support/commands/performanceMeasure/performanceMeasure.d.ts
   * I cannot find whether/where this is actually being used. And if it is, I _think_ it's just trying to interact with the screen from Cypress' perspective? So using different types from the underlying system may be perfectly appropriate.

Oddly, services/ui-src/src/utils/getLabelText.tsx appears to already be compatible with this change, despite not being touched in this branch. I can't wrap my head around this function either. It _probably_ still works as expected?

services/ui-src/src/measures/2023/shared/CommonQuestions/PerformanceMeasure/index.test.tsx
ailZhou: good catch, that was junk i forgot to remove

services/ui-src/src/measures/2023/shared/CommonQuestions/DeviationFromMeasureSpecification/index.tsx
ailZhou: this is an extremely good catch and i'm going to remove the sort. After testing, you were right, it does absolutely nothing. never did and never will.

services/ui-src/src/measures/2023/rateLabelText.ts
ailZhou: No more convention, they want it to be completely unique ids to prevent human error

services/ui-src/src/measures/2023/shared/globalValidations/validatePartialRateCompletion/index.ts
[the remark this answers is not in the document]
ailZhou: Good question, it's more of a check to see if there is a category being used or not so checking label is desired over checking id. After checking it will choose whether to pass the categories, if there is any, or pass an undefined for no categories

benmartin-coforma (dismissed):
LGTM!

BearHanded (dismissed):

benmartin-coforma (approved):

BearHanded (approved):

PR #1782, opened 2023-04-28
IU-HH Text Change 2382

benmartin-coforma:
### Description
Just a small text change for this measure

### Related ticket(s)
MDCT-2382

---
### How to test
1. Log in to a state with Health Home measures (for example, `stateuser6`).
2. Select FFY 2023.
3. Navigate to this measure.
4. Fill out the first three questions, to get the performance measure text to appear.
5. Check that text.

### Important updates
n/a

---
### Author checklist

- [-] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
- [-] I have updated relevant documentation, if necessary


ajaitasaini:
To clarify, you were getting rid of the text "health home" here?

keeysnc (approved):

services/ui-src/src/measures/2023/IUHH/data.ts
```diff
@@ -43,7 +43,7 @@ export const data: DataDrivenTypes.PerformanceMeasure = {
   customPrompt:
     "Enter the appropriate data below. Completion of at least one set of Numerator/Denominator/Rate (numeric entry, other than zero) is required.",
   questionText: [
-    "Rate of acute inpatient care and services (total, maternity, mental and behavioral disorders, surgery, and medicine) per 1,000 enrollee months among health home enrollees.",
+    "Rate of acute inpatient care and services (total, maternity, mental and behavioral disorders, surgery, and medicine) per 1,000 enrollee months among enrollees.",
```
ntsummers1, line 46: I think we want to keep the health home here? According to the ticket a red line just indicated a change. Not sure why its marked as a change though if the change is already good.. Possible we can just close this ticket and mark it as done but we should double check with David whether he actually wanted this to stay the same or be changed to something else

ntsummers1 (changes requested):
See above

benmartin-coforma:
@ajaitasaini @ntsummers1  You're totally right. I confirmed with David K that we don't want the text remove. I'm closing this PR and scooting the ticket across the board.

PR #1803, opened 2023-05-12
IET Total Rates / Validation

ailZhou:
### Description

For measure IET-AD & IET-HH, there's a new way of calculating total's in the N/D/R set. 

Normally, total summations are done in the qualifier layer, where the last qualifier is assumed to be the area for capturing the total data. 

In IET-AD & IET-HH, we want to capture data from the category layer, where the summation of the category's qualifiers is the sum of all the previous qualifiers of the same type (i.e. Ages 18 to 64 is the sum of all the previous Ages 18 to 64 in the other categories)

So it is sorted by categories: **Initiation of SUD Treatment** & **Engagement of SUD Treatment**.
Then further sorted by qualifiers: 
- Ages 18 to 64
- Age 65 and older

### Related ticket(s)

MDCT-2452

---
### How to test

1) Sign into QMR, any account
2) Select Reporting Year 2023
3) Go to IET-AD measure
4) Select first radio for first 3 questions
5) In N/D/R field, enter values in any category but the total categories ( Initiation of SUD Treatment: Total & Engagement of SUD Treatment: Total). 
6) Qualifiers in the Total categories should be the summation of all the previous non total categories.

### Important updates

---
### Author checklist

---


services/ui-src/src/components/IETRate/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+                numeratorSum.toString(),
+                denominatorSum.toString(),
+                rateMultiplicationValue,
+                1
+              )
+            : "0";
+      } else {
+        totalQual["rate"] = "";
+      }
+      // console.log(totalQual);
```
ntsummers1, line 244: Can remove this comment

services/ui-src/src/components/IETRate/index.tsx
```diff
@@ -0,0 +1,356 @@
...
+      if (
+        categoryName?.toLowerCase().includes("total") ||
+        prevRate[index]["label"].toLowerCase().includes("total")
+      ) {
+        prevRate[index]["isTotal"] = true;
+      }
+    });
+
+    field.onChange([...prevRate]);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
ntsummers1, line 103: Does this eslint still apply?

ntsummers1 (commented):
2 small things, but was able to pull this down and verify it works locally!

services/ui-src/src/components/IETRate/index.tsx
ailZhou: thanks for the catch, removed!

services/ui-src/src/components/IETRate/index.tsx
ailZhou: That's a really good question, I'm not sure as that is done for all the different kinds of rate components being utilized in QMR. I'm inclined to say it probably still applies as the only real change I made was expanding the conditional statement; the rest of the useEffect was a copy from the Rate component.

benmartin-coforma (approved):

BearHanded (dismissed):

keeysnc (approved):

BearHanded (approved):

PR #1817, opened 2023-05-31
MDCT-2337: Non-Required Measurement Fix (SPIKE Output)

karla-vm:
### Description

In a few instances, state users reported that the auto-completed (i.e. read-only) measures were being set to `In Progress` even though these should always be marked `Complete`. After taking a look with @BearHanded we saw that on a session time-out, the auto-completed measures are getting "saved" and thus returning to the `In Progress` state.

### Related ticket(s)

MDCT-2337

---
### How to test

- Navigate to any FY Adult Measures
- Verify that `NCIDDS-AD` is set to `Complete` by default
- Navigate away and back
- It should still be set `Complete`

### Important updates

---
### Author checklist

---


BearHanded (changes requested):
I think this bug would still break a normally completed measure if the user timed out while looking at it

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -224,7 +225,10 @@ export const MeasureWrapper = ({
   };
 
   const handleSave = (data: any) => {
-    if (!mutationRunning && !loadingData) {
+    // do not auto-save measure if the measure has already been completed or if it an auto-complete/read-only measure
+    const shouldSave =
+      measureStatus === MeasureStatus.INCOMPLETE || !autoCompletedMeasure;
+    if (!mutationRunning && !loadingData && !shouldSave) {
```
benmartin-coforma, line 231: I think this condition is backwards? Unless I'm misunderstanding, we want `shouldSave` and not `!shouldSave`.

karla-vm: That is correct, just fixed it! thanks 🏆

benmartin-coforma (dismissed):
Looks good!

BearHanded (approved):

keeysnc (approved):

PR #1831, opened 2023-06-14
Bug Fix: IU-HH Validation

ailZhou:
### Description

In IU-HH, the validation for comparing the number of enrollee months, which should be the same value across categories, is firing no matter if the value is the same or not. The error keeps showing **Value of "Number of Enrollee Months" in Ages 0 to 17 must be the same across all measurement categories.**, despite it being 0 to 17

### Related ticket(s)

MDCT-2620

---
### How to test

1) Sign into QMR, use an account that shows HH ([stateuserWA@test.com](mailto:stateuserDC@test.com))
2) Add Health Home measures
3) Go to IU-HH
4) Under Measure Specification section, select the first radio button
5) In Performance Measure, enter values into the N / D / R sets. The Number of Enrollee Months if different across categories should trigger the correct error per qualifier. 
- All ages 0 to 17 qualifiers should have the same Number of Enrollee Months or there will be a validation error
- Same applies for ages 18 to 64, age 65 and older & ages unknown

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/measures/2023/shared/globalValidations/ComplexValidations/ComplexValueSameCrossCategory/index.tsx
+++ b/services/ui-src/src/measures/2023/shared/globalValidations/ComplexValidations/ComplexValueSameCrossCategory/index.tsx
@@ -37,7 +37,6 @@ export const ComplexValueSameCrossCategoryOMS = (
     }
   }
 
-  // if (performanceMeasureArray)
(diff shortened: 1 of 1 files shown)


benmartin-coforma (approved):
Tested in Cloudfront. Got the expected error when a single age group had a different number of enrollee months for different services. I don't fully understand how this bug originated, or why this change fixes it, but I'm on board!

PR #1834, opened 2023-06-15
Add missing validation for CPU-AD

benmartin-coforma:
### Description
CPU-AD has two rates: plans with core documentation, and plans with supplemental documentation. These rates are talking about the same universe of plans, so the denominators should be the same. And we assume that any plan with supplemental documentation will have core documentation, so the first rate should be larger. These two facts are now part of the validation for the measure.

While I was working on this, I noticed that CPU-AD still has OMS validation in place, despite its OMS section having been recently removed. That dead code is now deleted.

### Related ticket(s)
MDCT-2624

---
### How to test
1. Open the CPU-AD measure for FFY 2023.
2. Enter a rate for Core, like 2/10
3. Enter a higher rate for Supplemental, with a different denominator. Like 4/8
4. Note the validation errors
5. Change the Supplemental rate to match denominators. Note that error goes away.
6. Change the Supplemental rate to be less than or equal to the Core rate. Note that the error goes away.

### Important updates
none

---
### Author checklist


karla-vm (approved):
💯

ailZhou (approved):
LGTM!

PR #1840, opened 2023-06-26
[QMR] Run jest on PR and not commit requirement

ajaitasaini:
### Related ticket(s)

MDCT-

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -8,6 +8,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v2
-      - uses: pre-commit/action@v2.0.2
-        with:
-          extra_args: --all-files
+      #- uses: pre-commit/action@v2.0.2
+      #with:
+      #extra_args: --all-files


benmartin-coforma:
I would sum it up like this:
* Git itself only cares about the hooks defined in `.git/hooks` directory.
* We use the [`pre-commit`](pre-commit.com) library in this project, to manage our pre-commit hooks. It reads from `pre-commit-config.yml` and overwrites files in `.git/hooks` every time you run `pre-commit install`.
* We ALSO have a bespoke script to manage hooks. It overwrites `.git/hooks` with the contents of `.githooks` every time you run `./dev`.
* Therefore, the hooks you execute locally depend on _which of those you ran most recently_.
* In this PR, Ajaita is removing the bespoke solution.

This has three effects:
1. Unit tests will no longer run on commit. They were included in the `.githooks` directory, but not in `pre-commit-config.yml`. This was the original goal of the story.
2. Prettified files will no longer be automatically staged. The `.githooks` script ran prettier _and re-staged all affected files_ before committing. That was an overreach that the `pre-commit` library plugin will not make. This change may surprise some developers, but it matches the behavior of other repos such as MCR.
3. Commits containing large files or secrets will fail. These checks have been included for a long time in `pre-commit-config.yml`, but doubtlessly they have been overwritten on most developers' machines by the bespoke hook management.

As Ajaita noted, it will be important for all developers to pull master and run `pre-commit install` once this PR is merged.

benmartin-coforma:
I do have one request. @braxex, @BearHanded: Would there be any ill effects from renaming `.github/workflows/pre-commit.yml` to something more descriptive, like `prettify-check.yml` or `pullrequest-precommit-check`? It is _not_ a pre-commit hook, but its existence & name did make this task more confusing.

If that's fine, I think this PR would be a good opportunity to make that change.

benmartin-coforma (approved):
Beautiful.

PR #1844, opened 2023-06-30
[QMR] MDCT-2548 Run jest on PR and not commit requirement

ajaitasaini:
### Description

pre-commit tests on QMR are taking forever, removing so they run on the PR per push

**prior to this change:** `./git/hooks/pre-commit` is responsible for running the script to run unit tests, and is _supposed_ to be populated by configurations set up in `pre-commit-config.yaml`. instead of `pre-commit-config.yaml` being utilized to run unit tests, they're instead being configured via a DIY pre-commit library under `.githooks/pre-commit` 

this comes into play when `./dev local` is run, specifically this command:

`# add/replace .githooks to .git/hooks
cp -a .githooks/. .git/hooks`

was causing `.git/hooks` to be populated by `.githooks/` instead of `pre-commit-config.yaml`, and hence running the custom command to run unit tests. since devs are more likely to be running `./dev local` far more than `pre-commit install`, is also the reason that prettier was likely not running correctly on most people's local dev environments 

**after this change:** since we no longer need to be running unit tests pre-commit, we can get rid of the library and the command to copy it over to `.git/hooks`, allowing `.git/hooks` to rely on `pre-commit-config.yaml` for prettier setup

### Related ticket(s)

MDCT-2548

---
### How to test

- pull changes in locally
- run `pre-commit install`
- confirm that `git commit` doesn't trigger tests and prettier works as expected (confirmed on my end)
- confirm that tests re-run on PR per push

### Important updates

- make sure to run `pre-commit install` in your local environment
- SUPER IMPORTANT: this PR will change the way files are staged r/e prettier
    - when you now attempt a commit, the pre-commit hook runs prettier
    - prettier will format the files and save them, but **will not stage them for you**
    - you can then review the files or stage them and re-commit

---
### Author checklist

---


karla-vm (approved):
🙏🏽

benmartin-coforma (dismissed):

benmartin-coforma:
I would sum it up like this:
* Git itself only cares about the hooks defined in `.git/hooks` directory.
* We use the [`pre-commit`](pre-commit.com) library in this project, to manage our pre-commit hooks. It reads from `pre-commit-config.yml` and overwrites files in `.git/hooks` every time you run `pre-commit install`.
* We ALSO have a bespoke script to manage hooks. It overwrites `.git/hooks` with the contents of `.githooks` every time you run `./dev`.
* Therefore, the hooks you execute locally depend on _which of those you ran most recently_.
* In this PR, Ajaita is removing the bespoke solution.

This has three effects:
1. Unit tests will no longer run on commit. They were included in the `.githooks` directory, but not in `pre-commit-config.yml`. This was the original goal of the story.
2. Prettified files will no longer be automatically staged. The `.githooks` script ran prettier _and re-staged all affected files_ before committing. That was an overreach that the `pre-commit` library plugin will not make. This change may surprise some developers, but it matches the behavior of other repos such as MCR.
3. Commits containing large files or secrets will fail. These checks have been included for a long time in `pre-commit-config.yml`, but doubtlessly they have been overwritten on most developers' machines by the bespoke hook management.

This PR was also a good opportunity to rename the "pre-commit" Github Action; it does not run before committing, so the name was confusing.

As Ajaita noted, it will be important for all developers to pull master and run `pre-commit install` once this PR is merged.

_(this comment copied from the original PR https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/1840 )_

benmartin-coforma (approved):

BearHanded (approved):

PR #1845, opened 2023-07-03
bugfix: ensure OMS checkbox displayValue is a string

braxex:
### Description

A bug is occurring when users attempt to preview the pdf on the `/pdf` route, which uses the `ExportAll` component. Nested deep inside the code for that part of the app was an issue where, if a user had entered any OMS data (thus requiring OMS data to be displayed on that page), an error would be thrown. This is because for OMS checkbox labels, where it looks for the label text to display, we were getting it 2 ways — a primary and a fallback. The problem was the specified label text was an `object` instead of a `string`, so the primary way was failing, then the backup way was grabbing the entire object where it was expecting just a string, and the app won’t render objects as children, and there you go: error.

Extent of this bug is limited to 2023 and OMS data when previewed on `/pdf` route.

Kudos to @benmartin-coforma @britt-mo @ajaitasaini for the great work diving in and researching everything, and @benmartin-coforma for identifying the solution.

Additional test coverage against this type of issue will be deployed with the change from e2e tests to 2023 coverage.

### Related ticket(s)

n/a

---
### How to test

1. For any core set, fill out a measure which includes OMS data. Save/complete measure.
2. Generate PDF preview by clicking 'Export', which will render `/pdf` route. If data is displayed, bug is fixed.

### Important updates

n/a

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/measures/2023/shared/CommonQuestions/OptionalMeasureStrat/ndrSets.tsx
+++ b/services/ui-src/src/measures/2023/shared/CommonQuestions/OptionalMeasureStrat/ndrSets.tsx
@@ -2,7 +2,7 @@ import * as CUI from "@chakra-ui/react";
 import * as DC from "dataConstants";
 import * as Types from "../types";
 import * as QMR from "components";
-import { LabelData, cleanString, getLabelText } from "utils";
+import { LabelData, cleanString } from "utils";
 import { useFormContext } from "react-hook-form";
 import { ComponentFlagType, usePerformanceMeasureContext } from "./context";
 import { useTotalAutoCalculation } from "./omsUtil";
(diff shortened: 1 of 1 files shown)


britt-mo (approved):

ajaitasaini (dismissed):

ajaitasaini (approved):

benmartin-coforma:
One small correction to Brax's explanation: The primary way to get label text was failing because it was path-dependent.

Specifically, on an individual measure's page, the url would look something like this: `.../MA/2023/ACS/AAB-AD`. The `2023` and `AAB-AD` would be extracted out, and we would look up the category labels based on year and measure name.

But the export page's url looks like this: `.../MA/2023/ACS/pdf`. We would attempt to look up the 2023 labels for measure `pdf`. Since that is not a measure name, no such labels exist, and we would use the fallback. And as Brax mentioned, that fallback in prior years was a string - but for 2023, it was a more sophisticated object.

Prior years still use this technique to get category labels. Since we use a more sophisticated category label/text/id system in 2023, we can rely on it completely - skipping the primary+fallback technique entirely.

PR #1878, opened 2023-07-13
[MDCT-2416] default scan to iterate across results

BearHanded:
### Description

Defaults the scan operation to continue when LastEvaluatedKey is present. This should allow complete deleteion of measures in lower environments, in addition to preventing future oversights when LastEvaluatedKey is ignored.

Also retained the original operation as a scanOnce() operation, as it can still be useful to break early, or in the case of forceKafkaSync, handle the results of a batch one page at a time.

### Related ticket(s)

MDCT-2416

---
### How to test

- Generate enough data locally to see the LastEvaluatedKey trigger which takes 1MB of data in a scan. One option for doing this is to find a large string (.5MB) to put in a free text area in the database directly. Do that a few times across items.
- All those results should still return in the UI when iterating through pages. You can do this by inflating a bunch of coreset measures and making sure all the results still appear.
- A delete operation on that same coreset should delete everything and not leave orphaned measures in the db

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/libs/dynamodb-lib.ts
+++ b/services/app-api/libs/dynamodb-lib.ts
@@ -37,8 +37,16 @@ export default {
   put: (params: DynamoCreate) => client.put(params).promise(),
   post: (params: DynamoCreate) => client.put(params).promise(),
   scan: async <Result = CoreSet | Measure>(params: DynamoScan) => {
-    const result = await client.scan(params).promise();
-    return { ...result, Items: result?.Items as Result[] | undefined };
+    const items = [];
+    let complete = false;
+    while (!complete) {
+      const result = await client.scan(params).promise();
+      console.log(result);
+      items.push(...(result?.Items as Result[] | []));
+      params.ExclusiveStartKey = result.LastEvaluatedKey;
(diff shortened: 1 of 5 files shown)


services/app-api/libs/dynamodb-lib.ts
```diff
@@ -36,7 +36,26 @@ export default {
...
   post: (params: DynamoCreate) => client.put(params).promise(),
+  /**
+   * Scan operation that continues for all results. More expensive but avoids stopping early when a index is not known.
+   */
   scan: async <Result = CoreSet | Measure>(params: DynamoScan) => {
+    const items = [];
+    let complete = false;
+    while (!complete) {
+      const result = await client.scan(params).promise();
+      console.log(result);
```
benmartin-coforma, line 47: Nit: We probably don't need the `console.log` here.

braxex (dismissed):
Didn't pull code down and test, but it all looks reasonable. Once Ben's comments are addressed, let's merge. 👍

braxex (approved):

benmartin-coforma (approved):

PR #1879, opened 2023-07-13
APM-CH/WCCCH - Validation Error

keeysnc:
### Description
For these validations, an `[object Object] ` was being displayed instead of the qualifier label name. The error was due to the parameter type being 'string' but the actual value being an 'object'. Since there is only 2 (APM-CH / WCCCH) measures that are displaying this issue I thought it would be a easier solution to ignore the type check on that specific input.

Before
Screenshot 2023-07-13 at 12 51 48 PM

After
Screenshot 2023-07-13 at 12 53 01 PM

### Related ticket(s)
https://qmacbis.atlassian.net/browse/MDCT-2683
MDCT-2683

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/measures/2023/APMCH/validation.ts
+++ b/services/ui-src/src/measures/2023/APMCH/validation.ts
@@ -31,7 +31,13 @@ const APMCHValidation = (data: FormData) => {
     qualifier: string,
     fieldType: string
   ) => {
-    return `${fieldType} for the ${qualifier} Total rate is not equal to the sum of the ${qualifier} age-specific ${fieldType.toLowerCase()}s.`;
+    return `${fieldType} for the ${
+      // @ts-ignore
+      qualifier.label
+    } Total rate is not equal to the sum of the ${
+      // @ts-ignore
+      qualifier.label
+    } age-specific ${fieldType.toLowerCase()}s.`;
   };
 
   errorArray = [
--- a/services/ui-src/src/measures/2023/APMCH/validation.ts
+++ b/services/ui-src/src/measures/2023/APMCH/validation.ts
@@ -31,13 +31,7 @@ const APMCHValidation = (data: FormData) => {
     qualifier: string,
     fieldType: string
   ) => {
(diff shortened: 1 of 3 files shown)


karla-vm (approved):

britt-mo (approved):

benmartin-coforma (changes requested):
This change would leave in place a bug for any future measures that may call `validateTotalNDR` with three arguments (using the default error message builder).

The root cause is an `any` on line 134 of `globalValidations/validateTotals/index.tsx`, in the declaration of `totalNDR`. If you remove that `any`, the typechecker will find the issues on lines 145 and 156: We will want `categories[0].label`, rather than just `categories[0]`. The same fix will need to be applied to the construction of `fieldLabel`, on line 166.

Note: This bug isn't _currently_ surfacing for any measures beyond APM-CH or WCC-CH, but that's by coincidence. Since all other measures using this validation do not supply a `categories` argument, those three lines fall through and use `totalNDR.label` instead.

keeysnc:
> This change would leave in place a bug for any future measures that may call `validateTotalNDR` with three arguments (using the default error message builder).
> 
> The root cause is an `any` on line 134 of `globalValidations/validateTotals/index.tsx`, in the declaration of `totalNDR`. If you remove that `any`, the typechecker will find the issues on lines 145 and 156: We will want `categories[0].label`, rather than just `categories[0]`. The same fix will need to be applied to the construction of `fieldLabel`, on line 166.
> 
> Note: This bug isn't _currently_ surfacing for any measures beyond APM-CH or WCC-CH, but that's by coincidence. Since all other measures using this validation do not supply a `categories` argument, those three lines fall through and use `totalNDR.label` instead.

@benmartin-coforma Updated. Thanks for the clarification.

benmartin-coforma (approved):
Looks great!

braxex (approved):
Didn't pull down and test but code looks reasonable. LGTM.

PR #1899, opened 2023-07-27
Set Login Year to 2023

ailZhou:
### Description

We're finally doing it, we're making the login default reporting year 2023 in QMR. 

Also, cypress test have been updated to select year 2021 so that it doesn't break everything.

### Related ticket(s)

MDCT-2764

---
### How to test

- Sign into QMR
- Look at the Reporting Year, it's now 2023

We will never have to make this mistake again.

### Important updates

N/A

---
### Author checklist

- [X] I have performed a self-review of my code
- [X] I have added [thorough](https://bit.ly/3zPrxuZ) tests, if necessary
---

### DIFF
--- a/services/ui-src/src/views/AdminHome/index.tsx
+++ b/services/ui-src/src/views/AdminHome/index.tsx
@@ -7,7 +7,7 @@ import { useFlags } from "launchdarkly-react-client-sdk";
 
 export const AdminHome = () => {
   const [locality, setLocality] = useState("AL");
-  const releaseYearByFlag = useFlags()
+  const releaseYearByFlag = useFlags()?.["release2023"]
     ? config.currentReportingYear
     : parseInt(config.currentReportingYear) - 1;
   const navigate = useNavigate();
--- a/services/ui-src/src/views/AdminHome/index.tsx
+++ b/services/ui-src/src/views/AdminHome/index.tsx
@@ -3,9 +3,13 @@ import * as CUI from "@chakra-ui/react";
 import { stateAbbreviations } from "utils/constants";
 import { useNavigate } from "react-router";
 import config from "config";
+import { useFlags } from "launchdarkly-react-client-sdk";
 
 export const AdminHome = () => {
   const [locality, setLocality] = useState("AL");
+  const releaseYearByFlag = useFlags()
+    ? config.currentReportingYear
+    : parseInt(config.currentReportingYear) - 1;
   const navigate = useNavigate();
   return (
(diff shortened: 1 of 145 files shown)


britt-mo (dismissed):

gmrabian (dismissed):
🙌 
one line change for the code and 142 lines for cypress 😭

tests/cypress/cypress/integration/a11y/ADDCHpage.spec.ts
```diff
@@ -1,6 +1,7 @@
 describe("ADD-CH Page 508 Compliance Test", () => {
   it("Check a11y on ADD-CH Page", () => {
     cy.login();
+    cy.selectYear("2021");
```
gmrabian, line 4: since we had to go through the effort of adding all these, would it be worth it to turn this into a variable so that in the future it's a one line change like the other config?

ailZhou: i don't think it would be worth it to do it for this branch, but i do think that it's a good idea to do when we start updating the cypress test to look at 2023.

karla-vm (approved):
FINALLY!

services/ui-src/src/views/AdminHome/index.tsx
```diff
@@ -3,9 +3,13 @@ import * as CUI from "@chakra-ui/react";
 import { stateAbbreviations } from "utils/constants";
 import { useNavigate } from "react-router";
 import config from "config";
+import { useFlags } from "launchdarkly-react-client-sdk";
 
 export const AdminHome = () => {
   const [locality, setLocality] = useState("AL");
+  const releaseYearByFlag = useFlags()
```
BearHanded, line 10: I think this should be referencing `useFlags()?.["release2023"]` and not just `useFlags()` before making any logical decisions

services/ui-src/src/views/Home/index.tsx
```diff
@@ -5,9 +5,13 @@ import * as CUI from "@chakra-ui/react";
 import "./index.module.scss";
 import * as QMR from "components";
 import { useUser } from "hooks/authHooks";
+import { useFlags } from "launchdarkly-react-client-sdk";
 
 export function Home() {
   const { userRole, userState } = useUser();
+  const releaseYearByFlag = useFlags()
```
BearHanded, line 12: same here

BearHanded (approved):
Looks good!

karla-vm (approved):

PR #1909, opened 2023-08-01
MDCT-2765 Pt. III: Add New Delivery Systems Validations

karla-vm:
### Description

Adding two (2) new validations for delivery systems and their nested selections to FFY2023 measures

### Related ticket(s)

MDCT-2765

---
### How to test

- Log in as state user
- Navigate to FY2023
- Do not select an option for `Which delivery systems are represented in the denominator?`, like so:
 
Screenshot 2023-08-01 at 10 36 07 AM

Validate the measure and you should see this error message:

Screenshot 2023-08-01 at 10 36 30 AM

- Next, select an option for that same question but don't select anything for the nested responses, like so:

Screenshot 2023-08-01 at 10 36 16 AM

Validate the measure and you should see this other error message:

Screenshot 2023-08-01 at 10 37 18 AM

### Important updates

N/A

---
### Author checklist

---


ailZhou (approved):
Tested in browser, LGTM!

PR #1911, opened 2023-08-01
[MDCT-2638] logout-redirect

BearHanded:
### Description

Create an endpoint to handle redirect on successful logout, and all the config that is required.

The endpoint exists outside the normal auth gates, so returning to it doesn't trap you in an endless loop in prod.

### Related ticket(s)

MDCT-2638

---
### Important updates

---
### Author checklist

---


services/ui-auth/serverless.yml
```diff
@@ -129,7 +129,9 @@ resources:
         DefaultRedirectURI: ${self:custom.application_endpoint_url}
         LogoutURLs:
           - ${self:custom.application_endpoint_url}
```
gmrabian, line 131: do we wanna keep the app endpoint and `localhost:3000/`?

gmrabian (approved):
lgtm

services/ui-auth/serverless.yml
BearHanded: In my mind I was keeping it so we could easily change it back with minor confusion if we wanted to, but if there's a strong opinion to change it I will

keeysnc (approved):
looks good!

PR #1915, opened 2023-08-04
Add UID to IU-HH & AIF-HH Field Data

ailZhou:
### Description

IU-HH & AIF-HH have unique fields compare to the other measures. So the chain is Category -> Qualifier -> Fields. 

The ask is to add an id for the field as to prevent any issues if the field label changes. 

We have checked, and they don't require us to do chaining of the id name, so we're going with just the short id for the uid. 

So the data should look like the image below, in the database.
uid

### Related ticket(s)

MDCT-2770

---
### How to test

1) Sign into QMR, one with Health Home (i.e. [stateuserWA@test.com](mailto:stateuserDC@test.com))
2) Go to AIF-HH
3) Select the first radio button for the first 3 questions, this will open up performance measures
4) Fill out data for performance measures & OMS
5) Spin up local dynamo db
6) Find AIF-HH 2023 data in local dynamo db, there should now be uid keys in the data

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/ComplexRate/index.tsx
+++ b/services/ui-src/src/components/ComplexRate/index.tsx
@@ -1,6 +1,6 @@
 import * as CUI from "@chakra-ui/react";
 import { useController, useFormContext } from "react-hook-form";
-import { allNumbers, xNumbersYDecimals } from "utils";
+import { allNumbers, LabelData, xNumbersYDecimals } from "utils";
 import * as QMR from "components";
 import objectPath from "object-path";
 import { useEffect, useLayoutEffect } from "react";
@@ -19,7 +19,7 @@ interface Props extends QMR.InputWrapperProps {
   calcTotal?: boolean;
   allowNumeratorGreaterThanDenominator?: boolean;
   categoryName?: string;
-  inputFieldNames?: any[];
+  inputFieldNames?: string[] | LabelData[];
   measureName?: string;
   ndrFormulas?: ndrFormula[];
 }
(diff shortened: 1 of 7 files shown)


services/ui-src/src/components/ComplexRate/index.tsx
```diff
@@ -19,7 +19,7 @@ interface Props extends QMR.InputWrapperProps {
   calcTotal?: boolean;
   allowNumeratorGreaterThanDenominator?: boolean;
   categoryName?: string;
-  inputFieldNames?: string[];
+  inputFieldNames?: any[];
```
keeysnc, line 22: Should this stay as a string array since the field name type is a string?

services/ui-src/src/components/ComplexRate/index.tsx
```diff
@@ -120,6 +120,16 @@ export const ComplexRate = ({
 
   // Quick reference list of all rate indices
   const rateLocations = ndrFormulas.map((ndr) => ndr.rate);
+  let inputFields: any[] = [];
```
keeysnc, line 123: Same question from the above comment.

services/ui-src/src/components/ComplexRate/index.tsx
```diff
@@ -120,6 +120,16 @@ export const ComplexRate = ({
...
+
+  //because this is a shared component, data pre-2023 will need to be slightly formatted to work
+  if (inputFieldNames.length > 0) {
```
keeysnc, line 126:
```suggestion
  if (inputFieldNames.length > 0 && typeof inputFieldNames[0] === "string") {
```

keeysnc (commented):
Looks good! Just had a few questions/suggestions.

services/ui-src/src/components/ComplexRate/index.tsx
ailZhou: i had it to any, because it could be either a string[] or a LabelData[], due to this being shared between years. Ben just   updated it though to be more type strong

services/ui-src/src/components/ComplexRate/index.tsx
ailZhou: same reason on earlier

services/ui-src/src/components/ComplexRate/index.tsx
ailZhou: talked to Nate on slack, not needed anymore

benmartin-coforma (approved):
I am now a co-author here, but I do approve this PR.

keeysnc (approved):

PR #1927, opened 2023-08-09
Global OMS Validation

ailZhou:
### Description

Currently for OMS, when selecting a sub-category and not selecting a label, it will generate a validation error. This is correct and expected.

The issue is when you have two sub-classification selected and only one of them is filled out. This causes the system to think both sub-classification had its N/D/R filled out, so no validation error is shown. 

This branch should fix that and will generate a validation error for the non-filled out sub-classification.

### Related ticket(s)

MDCT-2791

---
### How to test

1) Sign into QMR, any account
2) Go to AAB-AD
3) Select the first radio button for the first 3 questions, this will open up the Performance Measures
4) Go to Performance Measure, fill out the N/D/R set, every N/D/R filled will generate an equivalent check box for the OMS
5) The OMS section should be available now
6) Select the 'Race' checkbox
7) Check any two sub classification
8) Only fill out 1 sub-classification N/D/R set
9) Click Validate
10) There should be a validation error for the non-filled out N/D/R set.

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/measures/2023/shared/globalValidations/omsValidator/index.ts
+++ b/services/ui-src/src/measures/2023/shared/globalValidations/omsValidator/index.ts
@@ -84,6 +84,7 @@ const validateNDRs = (
 ) => {
   const isFilled: { [key: string]: boolean } = {};
   const isDeepFilled: { [key: string]: boolean } = {};
+  const isClassFilled: { [key: string]: boolean } = {};
   const errorArray: FormError[] = [];
   // validates top levels, ex: Race, Geography, Sex
   const validateTopLevelNode = (node: OMS.TopLevelOmsNode, label: string[]) => {
@@ -173,6 +174,7 @@ const validateNDRs = (
       ""
     );
     checkIsDeepFilled(locationReduced, rateData);
+    checkIsClassFilled(locationReduced, rateData);
   };
   //checks at least one ndr filled
   const checkNdrsFilled = (rateData: RateData) => {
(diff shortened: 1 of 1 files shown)


services/ui-src/src/measures/2023/shared/globalValidations/omsValidator/index.ts
```diff
@@ -84,6 +84,7 @@ const validateNDRs = (
 ) => {
   const isFilled: { [key: string]: boolean } = {};
   const isDeepFilled: { [key: string]: boolean } = {};
+  const isClassFilled: { [key: string]: boolean } = {};
```
BearHanded, line 87: Nit, feel free to ignore: When I read `isClassFilled` here I jump to an object's Class, not a qmr classification. Would it be better names along the lines of `checkIsClassificationFilled`, even if it is a mouthful?

BearHanded (dismissed):
Approved, with the one naming question

keeysnc (dismissed):
lgtm!

britt-mo:
These two scenarios still result in no validation errors:
Screenshot 2023-08-10 at 10 28 22 AM
Screenshot 2023-08-10 at 10 28 47 AM

britt-mo (changes requested):
Left screenshots

keeysnc (approved):
lgtm!

BearHanded (approved):

PR #1943, opened 2023-08-11
Cypress 2023 update a11y specs

gmrabian:
See base PR: https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/1928

### DIFF
--- a/tests/cypress/support/commands/commands.ts
+++ b/tests/cypress/support/commands/commands.ts
@@ -198,7 +198,7 @@ function terminalLog(violations) {
 
 // axe api documentation: https://www.deque.com/axe/core-documentation/api-documentation/
 Cypress.Commands.add("checkA11yOfPage", () => {
-  cy.wait(3000);
+  cy.wait(500);
   cy.injectAxe();
   cy.checkA11y(
     null,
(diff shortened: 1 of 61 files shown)


tests/cypress/support/commands/commands.ts
```diff
@@ -198,14 +198,18 @@ function terminalLog(violations) {
...
   cy.injectAxe();
   cy.checkA11y(
     null,
     {
       // @ts-ignore
       values: ["wcag2a", "wcag2aa"],
       includedImpacts: ["serious", "critical"], // options: "minor", "moderate", "serious", "critical"
+      rules: {
+        // ignoring link in text block because we use this on every page
+        "link-in-text-block": { enabled: false },
```
gmrabian, line 211: NOTE: our pages now include links in text blocks so I disabled this test because otherwise they'd all fail. I am following up with design to see if this is ok.

karla-vm (approved):

tests/cypress/support/commands/commands.ts
gmrabian: Design encouraged us to add underlines to the two places where it shows up, so that we are closer aligned with CMSDS. 
https://design.cms.gov/foundation/typography/links/?theme=core

That way we can re-enable this rule.

britt-mo (approved):

PR #1960, opened 2023-08-22
Calculate to Zero

ailZhou:
### Description

When entering data into N/D/R if numerator and denominator is 0, the rate does not auto calculate. and leaves an empty input field. 

Screenshot 2023-08-22 at 11 18 10 AM

This change will have it fill the rate out with a zero

Screenshot 2023-08-22 at 11 19 02 AM

---

**Note:** AAB-AD & AAB-CH calculates a little differently, so the rate value would be 100%.

Screenshot 2023-08-22 at 10 14 22 AM

### Related ticket(s)

MDCT-2838

---
### How to test

1) Sign into QMR, any account
2) Select any measure that has a Performance Measure, i.e. AMR-AD
3) In the **Measurement Specification** section, select the first radio button to enable the **Performance Measure** section
4) In the Performance Measure section, enter 0 for Numerator and Denominator. Rate should now auto calculate to 0.0.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/rateFormulas/index.ts
+++ b/services/ui-src/src/utils/rateFormulas/index.ts
@@ -13,7 +13,9 @@ export const defaultRateCalculation: RateFormula = (
 ) => {
   const floatNumerator = parseFloat(numerator);
   const floatDenominator = parseFloat(denominator);
-  const floatRate = floatNumerator / floatDenominator;
+  let floatRate = floatNumerator / floatDenominator;
+  //zero dividing by zero becomes a NaN when dividing, converting it back to zero. only numbers are entered in so no issues otherwise
+  floatRate = isNaN(floatRate) ? 0 : floatRate;
   const roundedRate = fixRounding(
     floatRate * rateMultiplicationValue,
     numbersAfterDecimal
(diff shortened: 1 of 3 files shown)


services/ui-src/src/utils/rateFormulas/index.ts
```diff
@@ -13,7 +13,9 @@ export const defaultRateCalculation: RateFormula = (
 ) => {
   const floatNumerator = parseFloat(numerator);
   const floatDenominator = parseFloat(denominator);
-  const floatRate = floatNumerator / floatDenominator;
+  let floatRate = floatNumerator / floatDenominator;
+  //zero dividing by zero becomes a NaN when dividing, converting it back to zero. only numbers are entered in so no issues otherwise
+  floatRate = isNaN(floatRate) ? 0 : floatRate;
```
benmartin-coforma, line 18: Suggestion: it may be clearer here to skip the calculation than to recover from `NaN` after the fact. Something like this:

```ts
if (floatDenominator === 0) {
  return (0).toFixed(numbersAfterDecimal);
}
let floatRate = floatNumerator / floatDenominator;
```

This also handles the case when numerator is nonzero, in which case the division gives `Infinity` instead of `NaN`.

benmartin-coforma (approved):
LGTM!

keeysnc (approved):
lgtm!

PR #1968, opened 2023-08-29
split out e2e tests to trigger after deploy succeeds

gmrabian:
### Description
This hopefully fixes our cypress issue in the PR updating them to 2023, but also applies to all PRs where the e2e tests should only run if deploy succeeds.

[GitHub Actions docs on workflow_run triggers](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run)
and [docs for concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#examples-using-concurrency-and-the-default-behavior)

### Related ticket(s)
MDCT-2545

---
### How to test
Unfortunately GitHub workflows only run when they are in the default branch (master). We cannot see this working until it's merged.

---
### Author checklist

---


.github/workflows/pull-request-e2e.yml
```diff
@@ -0,0 +1,187 @@
...
+      - completed
+
+permissions:
+  id-token: write
+  contents: read
+  actions: read
+  pull-requests: write
+
+jobs:
+  e2e-tests-init:
```
gmrabian, line 16: everything below (with one noted exception) is as it was in the previous file.

.github/workflows/pull-request-e2e.yml
```diff
@@ -0,0 +1,187 @@
...
+    name: Initialize End To End Tests
+    if: ${{ github.ref != 'refs/heads/prod' && github.ref != 'refs/heads/val' && github.event.workflow_run.conclusion == 'success' }}
```
gmrabian, line 18: this is the difference: this if also checks to make sure Deploy not only finished, but succeeded. That way these only run if it actually deployed. Since all jobs after this depend on this job via `need` they will not run if this does not meet its conditions

.github/workflows/pull-request-e2e.yml
```diff
@@ -0,0 +1,187 @@
+name: PR E2E Tests
+
+on:
+  workflow_run:
+    workflows: [Deploy]
+    types:
+      - completed
```
gmrabian, line 7: instead of triggering on PR this now triggers on `Deploy` action completing, to make sure we're not running these before that happens

.github/workflows/pull-request-e2e.yml
```diff
@@ -0,0 +1,191 @@
...
+permissions:
+  id-token: write
+  contents: read
+  actions: read
+  pull-requests: write
+
+jobs:
+  e2e-tests-init:
+    name: Initialize End To End Tests
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
gmrabian, line 22: modified this if to only trigger this step if the `Deploy` succeeds. All steps after this one are dependent on this one, and will not run unless this one does.

.github/workflows/pull-request-e2e.yml
```diff
@@ -0,0 +1,191 @@
...
+
+on:
+  workflow_run:
+    workflows: [Deploy]
+    branches-ignore:
+      - prod
+      - val
+      - master
+    types:
+      - completed
```
gmrabian, line 11: Modified condition such that this workflow runs only when Deploy workflow completes. Additionally it ignores runs on prod, val, and master since we don't run e2e tests on those.

.github/workflows/deploy.yml
```diff
@@ -7,6 +7,9 @@ on:
       - "dependabot/**"
       - "!skipci*"
 
+concurrency:
+  group: ${{ github.ref }}
```
gmrabian, line 11: use native `concurrency` instead of branch lock script. only runs deploy if there are no previous deploys running on the same group (in this case, the ref, which contains the branch name)

.github/workflows/deploy.yml
```diff
@@ -7,6 +7,9 @@ on:
       - "dependabot/**"
       - "!skipci*"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
gmrabian, line 11: use native `concurrency` instead of branch lock script. only runs deploy if there are no previous deploys running on the same group (in this case, the ref, which contains the branch name)

BearHanded: does it get in line or just fail? The lock waits indefinitely I think

BearHanded (approved):

.github/workflows/deploy.yml
gmrabian: Docs make it sound like it'll wait indefinitely just like the lock

> When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending. Any previously pending job or workflow in the concurrency group will be canceled.

PR #1970, opened 2023-08-30
OMS Section Not Clearing

ailZhou:
### Description

OMS data isn't being cleared when switching between Measurement Specification types, i.e. HEDIS year and Other. 

If you enter Other OMS N/D/R first and validate, the data will be the same when looking at Hedis year OMS. 

https://github.com/Enterprise-CMCS/macpro-mdct-qmr/assets/127151429/a33aba69-7246-42a8-9df7-18f10ec6cc71

**Longer Explanation of Problem**
Other OMS and OMS are saved as part of the same data chunk, which means the data does not really differentiate between the two. The frontend parses that data based on ids but it also doesn't really know that the data belongs to two different sections. For example, if the user checkboxes Race and saves it, that data is just retrieved as Race and as long as there's a checkbox in the OMS section that matches, it will be automatically checked. It is why switching back and forth between the two section will trigger the Race checkbox to be checked; the checkbox data is decoupled from the section it is in.

**Explanation of Solution**
Due to the limited time until the go live date, I'm choosing to do a hard clear of the data when switching between the two sections. It is not the most ideal fix but it will do the job.

### Related ticket(s)

MDCT-2865

---
### How to test

1) Sign into QMR, any account
2) Go to any measure with an OMS, i.e AAB-AD
3) In Measure Specification section, select Other, this will open the Other Performance Measure section
4) Fill out an Other Performance Measure, this will trigger the OMS section to show up
5) Scroll down and fill out OMS
6) Validate
7) Switch to a Hedis Year in Measure Specification
8) Fill out Performance Measure
9) Scroll down to OMS, and you should be able to fill out OMS data normally.
10) If you switch back, Other section OMS data will be cleared

### Important updates

N/A

---
### Author checklist

---


gmrabian (approved):
works as described!

love the comments 😅

BearHanded (approved):

PR #1974, opened 2023-09-08
Fix roles

braxex:
#### Preamble
This is gonna be a long one. Hang in there. First, the main players:

- _ADMIN — This is the super admin or Business Owner Rep role. It is one of four "admin-type" (think: non-state-user) roles, and is the main role referred to as "Admin". It is held by only the application owner at the MDCT level and typically only one user at a time will have this role. Generally read-only, with the exception of the admin banner._
- _APPROVER — This is the approver user role, and it is one of four "admin-type" (think: non-state-user) roles. It is held by the lead Business Owners at CMS HQ with responsibility over QMR. Generally read only._
- _HELP DESK — This is the help desk user role, and it is one of four "admin-type" (think: non-state-user) roles.  It is held by help desk staff within MACPro assigned to Tier 2 support for the MDCT applications. Generally read only._
- _INTERNAL — This is the internal user role, and it is one of four "admin-type" (think: non-state-user) roles. It is held by internal CMS staff and contractors, like Mathematica and Coforma. Generally read only._
- _STATE USER — This is the state user role, and users with this type also have a state attribute. (Note: Any user can have a state attribute, but these ones need it.) It's held by state-level CMS staff and contractors. Has read/write access, but only for their associated state and not for the admin banner._

#### Chapter 1: An Unexpected Party
During validation of the new internal user role, I noticed that they had access to the `/admin` route and made a note to investigate further, assuming that we were simply not blocking that route in the UI & when routes were created, and that the API was surely locked down appropriately. I made a note to investigate further and thought nothing more of the matter. 

#### Chapter 2: A Shortcut to Madness
When I had more time to devote to the subject, I sat down to create what I thought would be a simple dozen-line PR adding a UI change and a matching test, only to find that we were not actually blocking the API and that it was, in fact, explicitly exposed to all users via an override prop being passed to the handler. I immediately set to work rectifying the situation, pulling ever so gently on the thread which would eventually unravel the entire mystery, but which first would lead me through the unknowable, unnamed parts of QMR.

#### Chapter 3: The Shadow of the Past
All API calls are wrapped in a [nifty handler](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/services/app-api/libs/handler-lib.ts#L11-L33) provided by an internal library, `lib-handler`. But it has a dark secret: an optional argument, `postOverride`, which is passed through to an authorization utility used to determine if the user is authorized to access resources. This [auth utility](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/services/app-api/libs/authorization.ts#L13-L41) was then using that override arg to determine if it should override certain checks and allow users to access resources regardless. I looked and it was only really being used by 3 handlers: `banners/create (createBanner)`, `banners/delete (deleteBanner)`, and `prince/pdf (getPdf)`. It was originally put in place to allow all users to interact with Prince XML, the pdf printing resource, because although the method is named "getPdf", it's a POST, and prior to it's adoption, QMR was only allowing state users to POST (admin-type users had no reason to do so back in these days). The implementation of the admin banner used this as well to allow admin users to POST and DELETE banners. But this inadvertently opened up the ability for anyone to do so, because as it turns out, the authorization util was doing a lot more than it should have been.

#### Chapter 4: A Utility Unmasked
The auth utility was not only checking for authentication, but also performing two other functions: (1) Checking for state permissions, and (2) Checking for role permissions. This would normally be ok, but the combination of the two was causing some problems. Some pseudocode of what it was doing for illustration (or you can look at the code yourself [here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/services/app-api/libs/authorization.ts#L30-L40)):

```js
const shouldWeAllowTheUserToDoThis = (event, postOverride) => {

// check if state user has the same state as the event (API call)
  if (isStateUser && userHasAssociatedState && EventParametersIncludesState) {
    return usersState === eventsState
  }

  // you got here, so user must be an admin*
  return (
    requestMethodIsGET ||                    // let everyone get stuff!
    requestMethodIsPOST && postOverride ||   // let the people post!
    requestMethodIsDELETE && postOverride   // let it all burn!
  );
}

// *eagle-eyed reviewers will notice this is not true. if the event doesn't include specify a state or if the state user doesn't have an associated state, you get here too.
```

#### Chapter 5: In The House of Brax Bombadil
Anyway, it was doing too much. And that is generally a bad™ thing. So I set to work fixing it. This wasn't too bad. I cleaned up `isAuthorized` (now renamed `isAuthenticated` thanks to @gmrabian), set up separate utilities in the `authorization.ts` util file to handle `hasStatePermissions` and `hasRolePermissions`, got those implemented and started writing up some tests. But testing started going bad. I was sure I was mocking things right. Or was I? Was it all a dream? _Did_ I used to read _Word Up!_ magazine? I had to find the answer, and quick. You see, I was expecting a series of 403 status codes to come back in tests, because I was testing that unauthorized users would get rejected. Instead, I kept getting 200s. Why?! The more I looked into it, the more positive I was that something odd was afoot. Then I figured it out. The response from each API call was something like this:

```js
const response = await dynamoDb.post(params);
console.log(response);

/*
{
  statusCode: 200,
  body: {
    statusCode: 403,
    body: "User is not authorized blah blah blah..."
  }
}
*/
```

You can imagine my surprise. 

#### Chapter 6: The Old Forest
As it turns out, another internal library, `response-lib.ts`, had a method called `success`, which was being used in that wrapping handler to just always return success (unless a user wasn't authenticated), making it so that any other failure, short of throwing an error and crashing, would just silently fail and return you that failure wrapped in a neat little 200 success package. Fun, huh?

#### Chapter 7: At the Sign of the Prancing Brax
Once I figure this out, things were seeming easier. Sure I had to change all the tests that were relying on these naive 200s coming back, but that was a small price for a clean API that returns what it means to, right? Right? Tell me it's worth it, please.

#### Chapter 8: Fog on the Braxow-Downs
By this point, I was feeling like I'd rewritten half the backend of QMR and was starting to feel kinda nervous. I hadn't even gotten to my original goal yet! (You'll remember I started this story approximately 1 year ago, and had planned to just hide the admin banner button from non super-admins and make sure they couldn't hit that route). So off I went to go figure out that stuff. But oh yeah, the user roles were all weird. So I labeled them. And I was like, wait, what is a `mdctqmr-bo-user`? Isn't it just `mdctqmr-bor`? I dug up the old QMR IDM tech specs and could find no mention of this BO User. I checked all the deployed AWS environment Cognito user pools, and likewise, not a trace of any user who had this BO User role. You'll note it is now removed. But I was real confused for a bit in there.

#### Chapter 9: Coder / Write in the Dark
Now I was really cruising. I cleaned up the user types. I cleaned up the users.json, I cleaned up the UI, I hid the button, I changed the routes. I was doing it! It was really happening! I made the banner appear on the AdminHome page and the user scroll up after submitting the banner. UX improvements! I checked all the existing instances of user role stuffs. I wrote comments! Such comments!

#### Chapter 10: Flight to the PR
And now we're here. Changes have been made and it's time to review them. I thank you for your time. I thank you for your diligence. If you read this, I thank you for your patience. Let's talk about what changed.

### Changes Made
- Modified handler authorization permissions (by calling new helper utilities) as follows:
    - Banners
        - createBanner: restricted to ADMIN
        - deleteBanner: restricted to ADMIN
        - fetchBanner: no auth changes, remains open to all users
    - Core Sets
        - createCoreSet: restricted to Admin* & STATE USERS (with matching state)
        - coreSetList: restricted to Admin* & STATE USERS (with matching state)
        - getCoreSet: restricted to Admin* & STATE USERS (with matching state)
        - deleteCoreSet: restricted to STATE USERS (with matching state)
        - editCoreSet: restricted to STATE USERS (with matching state)
    - Measures
        - createMeasure: restricted to Admin* & STATE USERS (with matching state)
        - listMeasure: restricted to Admin* & STATE USERS (with matching state)
        - getMeasure: restricted to Admin* & STATE USERS (with matching state)
        - deleteMeasure: restricted to STATE USERS (with matching state)
        - editMeasure: restricted to STATE USERS (with matching state)
    - Prince PDF: no auth changes, remains open to all users
- Modified handler-lib handler to no longer allow override of authorization checks
- Modified authorization lib to separate authentication concerns from permission concerns
    - Changed isAuthorized to isAuthenticated and simplified, moving checks to other helpers
    - Added hasStatePermissions and hasRolePermissions utils
- Modified response-lib handler to no longer default to success response that obfuscates true response in nested object
- Cleaned up user role types (ADMIN, APPROVER, HELP DESK, INTERNAL, STATE USER)
    - Removed unused BO user role
- Cleaned up references to user roles
- Updated all related tests and wrote new ones
- Oh yeah changed some UI stuff to block non-admin-type users from getting to `/admin` and block non-super-admin users from getting to `/admin/banner` and seeing the admin banner button. Plus some QOL improvements for the super admin UX.

### Related ticket(s)

n/a

---
### How to test

We should test each of the following handler & user combinations in the application UI to ensure they are working as expected (✅ = should work, 🚫 = should not work). *Admin-Type = APPROVER, HELP DESK, or INTERNAL

#### Banners
| # | File | Handler | ADMIN | Non Super Admin | STATE  | (Right State) | (Wrong State)
| :--------:  | ------- | ------- | :-------: | :-------: | :-------: | :-------: | :-------: | 
| 1 | banners/create | createBanner | ✅ | 🚫 | 🚫 | — | — | 
| 2 | banners/delete | deleteBanner | ✅ | 🚫 | 🚫 | — | — | 
| 3 | banners/fetch | fetchBanner | ✅ | ✅ | ✅ | — | — | 

#### Core Sets
| # | File | Handler | ADMIN | Non Super Admin | STATE  | (Right State) | (Wrong State)
| :--------:  | ------- | ------- | :-------: | :-------: | :-------: | :-------: | :-------: | 
| 4 | coreSets/create | createCoreSet | ✅ | ✅ | → | ✅ | 🚫 |
| 5 | coreSets/delete | deleteCoreSet | 🚫 | 🚫 | → | ✅ | 🚫 |
| 6 | coreSets/get | coreSetList | ✅ | ✅ | → | ✅ | 🚫 |
| 7 | coreSets/get | getCoreSet | ✅ | ✅ | → | ✅ | 🚫 |
| 8 | coreSets/update | editCoreSet | 🚫 | 🚫 | → | ✅ | 🚫 |

#### Measures
| # | File | Handler | ADMIN | Non Super Admin | STATE  | (Right State) | (Wrong State)
| :--------:  | ------- | ------- | :-------: | :-------: | :-------: | :-------: | :-------: | 
| 9 | measures/create | createMeasure | ✅ | ✅ | → | ✅ | 🚫 |
| 10 | measures/delete | deleteMeasure | 🚫 | 🚫 | → | ✅ | 🚫 |
| 11 | measures/get | getMeasure | ✅ | ✅ | → | ✅ | 🚫 |
| 12 | measures/get | listMeasures | ✅ | ✅ | → | ✅ | 🚫 |
| 13 | measures/update | editMeasure | 🚫 | 🚫 | → | ✅ | 🚫 |

#### Other
| # | File | Handler | ADMIN | Non Super Admin | STATE  | (Right State) | (Wrong State)
| :--------:  | ------- | ------- | :-------: | :-------: | :-------: | :-------: | :-------: | 
| 14 | prince/pdf | getPdf | ✅ | ✅ | ✅ | — | — | 

### Important updates

n/a

---
### Author checklist

---

### DIFF
--- a/services/app-api/handlers/coreSets/update.ts
+++ b/services/app-api/handlers/coreSets/update.ts
@@ -41,6 +41,5 @@ export const editCoreSet = handler(async (event, context) => {
     ),
   };
   await dynamoDb.update(params);
-
-  return params;
+  return { status: StatusCodes.SUCCESS, body: params };
 });
--- a/services/app-api/handlers/coreSets/update.ts
+++ b/services/app-api/handlers/coreSets/update.ts
@@ -2,9 +2,25 @@ import handler from "../../libs/handler-lib";
 import dynamoDb from "../../libs/dynamodb-lib";
 import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpressionVars";
 import { createCompoundKey } from "../dynamoUtils/createCompoundKey";
-import { getUserNameFromJwt } from "../../libs/authorization";
+import {
+  getUserNameFromJwt,
+  hasRolePermissions,
+  hasStatePermissions,
+} from "../../libs/authorization";
+import { UserRoles } from "../../types";
+import { Errors, StatusCodes } from "../../utils/constants/constants";
 
 export const editCoreSet = handler(async (event, context) => {
+  // action limited to state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  const isFromCorrespondingState = hasStatePermissions(event);
+  if (!isStateUser || !isFromCorrespondingState) {
+    return {
+      status: StatusCodes.UNAUTHORIZED,
+      body: Errors.UNAUTHORIZED,
+    };
+  }
+
   const { submitted, status } = JSON.parse(event!.body!);
   const dynamoKey = createCompoundKey(event);
   const lastAlteredBy = getUserNameFromJwt(event);
--- a/services/app-api/handlers/coreSets/update.ts
+++ b/services/app-api/handlers/coreSets/update.ts
@@ -4,17 +4,13 @@ import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpress
 import { createCompoundKey } from "../dynamoUtils/createCompoundKey";
 import {
   getUserNameFromJwt,
-  hasRolePermissions,
   hasStatePermissions,
 } from "../../libs/authorization";
-import { UserRoles } from "../../types";
 import { Errors, StatusCodes } from "../../utils/constants/constants";
 
 export const editCoreSet = handler(async (event, context) => {
   // action limited to state users from corresponding state
-  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
-  const isFromCorrespondingState = hasStatePermissions(event);
-  if (!isStateUser || !isFromCorrespondingState) {
+  if (!hasStatePermissions(event)) {
     return {
       status: StatusCodes.UNAUTHORIZED,
       body: Errors.UNAUTHORIZED,
--- a/services/app-api/handlers/measures/get.ts
+++ b/services/app-api/handlers/measures/get.ts
@@ -86,9 +86,15 @@ export const getMeasure = handler(async (event, context) => {
 
 export const getReportingYears = handler(async () => {
   const reportingYears = Object.keys(measures);
-  return reportingYears;
+  return {
+    status: StatusCodes.SUCCESS,
+    body: reportingYears,
+  };
 });
 
 export const getMeasureListInfo = handler(async () => {
-  return measures;
+  return {
+    status: StatusCodes.SUCCESS,
+    body: measures,
+  };
 });
--- a/services/app-api/handlers/measures/get.ts
+++ b/services/app-api/handlers/measures/get.ts
@@ -3,10 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
 import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpressionVars";
 import { createCompoundKey } from "../dynamoUtils/createCompoundKey";
 import { measures } from "../dynamoUtils/measureList";
-import { Measure } from "../../types";
+import {
+  hasRolePermissions,
+  hasStatePermissions,
+} from "../../libs/authorization";
+import { Measure, UserRoles } from "../../types";
+import { Errors, StatusCodes } from "../../utils/constants/constants";
 import { Key } from "aws-sdk/clients/dynamodb";
 
 export const listMeasures = handler(async (event, context) => {
+  // action limited to any admin type user and state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  if (isStateUser) {
+    const isFromCorrespondingState = hasStatePermissions(event);
+    if (!isFromCorrespondingState) {
+      return {
+        status: StatusCodes.UNAUTHORIZED,
+        body: Errors.UNAUTHORIZED,
+      };
+    }
+  } // if not state user, can safely assume admin type user due to baseline handler protections
+
   const state = event.pathParameters?.state;
   const year = event.pathParameters?.year as string;
   const coreSet = event.pathParameters?.coreSet;
@@ -37,6 +54,18 @@ export const listMeasures = handler(async (event, context) => {
 });
 
 export const getMeasure = handler(async (event, context) => {
+  // action limited to any admin type user and state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  if (isStateUser) {
+    const isFromCorrespondingState = hasStatePermissions(event);
+    if (!isFromCorrespondingState) {
+      return {
+        status: StatusCodes.UNAUTHORIZED,
+        body: Errors.UNAUTHORIZED,
+      };
+    }
+  } // if not state user, can safely assume admin type user due to baseline handler protections
+
   const dynamoKey = createCompoundKey(event);
   const params = {
     TableName: process.env.measureTableName!,
@@ -47,13 +76,4 @@ export const getMeasure = handler(async (event, context) => {
   };
   const queryValue = await dynamoDb.get(params);
   return queryValue;
-});
-
-export const getReportingYears = handler(async () => {
-  const reportingYears = Object.keys(measures);
-  return reportingYears;
-});
-
-export const getMeasureListInfo = handler(async () => {
-  return measures;
 });
--- a/services/app-api/handlers/measures/get.ts
+++ b/services/app-api/handlers/measures/get.ts
@@ -83,3 +83,12 @@ export const getMeasure = handler(async (event, context) => {
     body: queryValue,
   };
 });
+
+export const getReportingYears = handler(async () => {
+  const reportingYears = Object.keys(measures);
+  return reportingYears;
+});
+
+export const getMeasureListInfo = handler(async () => {
+  return measures;
+});
--- a/services/app-api/handlers/measures/get.ts
+++ b/services/app-api/handlers/measures/get.ts
@@ -50,7 +50,10 @@ export const listMeasures = handler(async (event, context) => {
     });
   });
   queryValue.Items = scannedResults;
-  return queryValue;
+  return {
+    status: StatusCodes.SUCCESS,
+    body: queryValue,
+  };
 });
 
 export const getMeasure = handler(async (event, context) => {
@@ -75,5 +78,8 @@ export const getMeasure = handler(async (event, context) => {
     },
   };
   const queryValue = await dynamoDb.get(params);
-  return queryValue;
+  return {
+    status: StatusCodes.SUCCESS,
+    body: queryValue,
+  };
 });
--- a/services/app-api/handlers/measures/update.ts
+++ b/services/app-api/handlers/measures/update.ts
@@ -4,17 +4,13 @@ import { convertToDynamoExpression } from "../dynamoUtils/convertToDynamoExpress
 import { createCompoundKey } from "../dynamoUtils/createCompoundKey";
 import {
   getUserNameFromJwt,
-  hasRolePermissions,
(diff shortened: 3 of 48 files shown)


benmartin-coforma (commented):
I had a bunch of questions about this, mostly from my own curiosity. I threw them at Brax in a side channel, so I'm just leaving this comment as a placeholder.

services/app-api/libs/authorization.ts
```diff
@@ -10,34 +11,60 @@ interface DecodedToken {
   identities?: [{ userId?: string }];
 }
 
-export const isAuthorized = (
+export const isAuthorized = (event: APIGatewayProxyEvent) => {
```
gmrabian, line 14: realizing this should probably be `isAuthenticated` but I know we have it as `isAuthorized` in other apps (my fault) so I'm ok with consistency

services/app-api/handlers/measures/update.ts
```diff
@@ -2,9 +2,25 @@ import handler from "../../libs/handler-lib";
...
+  hasRolePermissions,
+  hasStatePermissions,
+} from "../../libs/authorization";
+import { UserRoles } from "../../types";
+import { Errors, StatusCodes } from "../../utils/constants/constants";
 
 export const editMeasure = handler(async (event, context) => {
+  // action limited to state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  const isFromCorrespondingState = hasStatePermissions(event);
```
gmrabian, line 16: `hasStatePermissions` technically checks `isStateUser` as well, right? Could we just use `hasStatePermissions` here?

services/app-api/handlers/measures/get.ts
```diff
@@ -33,10 +50,25 @@ export const listMeasures = handler(async (event, context) => {
...
+    status: StatusCodes.SUCCESS,
+    body: queryValue,
+  };
 });
 
 export const getMeasure = handler(async (event, context) => {
+  // action limited to any admin type user and state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  if (isStateUser) {
+    const isFromCorrespondingState = hasStatePermissions(event);
```
gmrabian, line 63: similar comment: could we just use `hasStatePermissions`?

services/app-api/handlers/measures/get.ts
```diff
@@ -3,10 +3,27 @@ import dynamoDb from "../../libs/dynamodb-lib";
...
+} from "../../libs/authorization";
+import { Measure, UserRoles } from "../../types";
+import { Errors, StatusCodes } from "../../utils/constants/constants";
 import { Key } from "aws-sdk/clients/dynamodb";
 
 export const listMeasures = handler(async (event, context) => {
+  // action limited to any admin type user and state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  if (isStateUser) {
+    const isFromCorrespondingState = hasStatePermissions(event);
```
gmrabian, line 18: etc.

services/app-api/handlers/coreSets/update.ts
```diff
@@ -2,9 +2,25 @@ import handler from "../../libs/handler-lib";
...
 export const editCoreSet = handler(async (event, context) => {
+  // action limited to state users from corresponding state
+  const isStateUser = hasRolePermissions(event, [UserRoles.STATE_USER]);
+  const isFromCorrespondingState = hasStatePermissions(event);
+  if (!isStateUser || !isFromCorrespondingState) {
+    return {
+      status: StatusCodes.UNAUTHORIZED,
+      body: Errors.UNAUTHORIZED,
+    };
+  }
```
gmrabian, line 22: if we keep this, does seem like it's a reusable pattern across many handlers

gmrabian (commented):
awesome security improvements here! Some suggestions for refactoring but I'm happy as is too

services/app-api/libs/authorization.ts
braxex: Changing.

benmartin-coforma:
JWTs by themselves are only encoded and not encrypted. @BearHanded , @gmrabian - Do you know if our JWTs are signed or otherwise verified on the API side before this code runs?

services/app-api/handlers/measures/update.ts
braxex: Yep it does. There's a little bit of nuance on some of these but for ones like this we can swap 'em. Good idea.

services/app-api/handlers/measures/get.ts
braxex: These ones are a little bit different because unlike the ones you noted before (like measures/update.ts) we allow admin-type users. So the `hasStatePermissions` check will always fail for admin users if we just use that one because they aren't state users, and we want to allow all admins, as well as state user who are from that specific state. If we try to refactor the same way with these ones we still end up having to do a check for user role type in some way, like:

```
if (isStateUser && !hasStatePermissions) throw Error;
or 
if (!isAdminUser && !hasStatePermissions) throw Error;
```

We could do either of those, but that's more or less what we're already doing, just with slightly different syntax. 🤷

BearHanded:
> JWTs by themselves are only encoded and not encrypted. @BearHanded , @gmrabian - Do you know if our JWTs are signed or otherwise verified on the API side before this code runs?

@benmartin-coforma I believe they are signed and verified between the amplify lib <->  cognito/api gateway

BearHanded (dismissed):
Matrix of permissions seems to behave as expected

BearHanded (approved):

benmartin-coforma (approved):
Code looks good to me.

gmrabian (approved):
probs fine until cypress finds us in our sleep 🔪

P.S. merge [kebab change](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/1975) first thank you

PR #2003, opened 2023-10-30
Add x-frame-options header

benmartin-coforma:
### Description
This PR adds an additional HTTP header when serving the website, which instructs browsers not to serve our content within an iframe.

Previously, the function which attaches HTTP headers for this app was named for the only header it cared about. It has been renamed appropriately.

### Related ticket(s)
CMDCT-271

---
### How to test
We don't use iframes to serve this website, so this PR shouldn't change anything. A quick smoke test (as provided by Cypress) should be sufficient.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -202,7 +202,7 @@ resources:
             ViewerProtocolPolicy: redirect-to-https
             FunctionAssociations:
               - EventType: viewer-response
-                FunctionARN: !GetAtt HeadersCloudfrontFunction.FunctionMetadata.FunctionARN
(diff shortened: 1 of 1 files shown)


karla-vm (dismissed):

gmrabian (commented):
This deployed ok so hopefully there are no issues. Did you try deploying a branch without the cloudfront function name change and then redeploying after? Just want to make sure there's no error when renaming an existing function

benmartin-coforma:
This is ready for re-review. Without the rename, the merge should go smoothly.

gmrabian (approved):

karla-vm (approved):

PR #2011, opened 2023-12-07
Othermeasure rates bug

gmrabian:
### Description

Other Performance Measure was crashing the render when you deleted a measure and saved because a null value got into the array. This fixes that issue. See video and How to Test for more info.

### Related ticket(s)

CMDCT-3062

---
### How to test

On this branch:
- Sign in as a state user
- Generate Child core sets separately
- Go to Child core sets: medicaid and go to DEV-CH (note: fix applies to other measures. This is just what we've been testing)
- Select `Yes`, `Final Data`, `Other`, `Other` for the first four options
- Scroll down to Other Performance Measures and create three (description not necessary)
- Click save after creating three
- Delete the middle measure and save again
- Edit any other field and save once more. 

If you run this on `master`
- The render should crash.
- In dynamodb-admin you can see the `null` value that causes the break. If you delete the `null` it goes away.

video of crash:
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/assets/57802560/5438c1fa-35b7-4425-8945-df1ffa3fcbbb

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -57,7 +57,7 @@ interface MeasureProps {
 }
 
 const Measure = ({ measure, handleSave, ...rest }: MeasureProps) => {
-  const { watch } = useFormContext();
+  const { watch, setValue } = useFormContext();
 
   const watchedData = useWatch();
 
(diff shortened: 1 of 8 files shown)


BearHanded (approved):
Looks good, unable to reproduce it on the branch, and I can reproduce it in main

gmrabian:
I created the PR, though Berry did all the work. I'm approving symbolically 👍

berryd (approved):

ailZhou (approved):
Tested in browser, data looks good. Thank you for fixing this @berryd!

PR #2036, opened 2024-01-18
Making QMR Dev Environments Not Publicly Accessible

dwhitestratiform:
### Description
This pull request is to lock down non production environments for QMR to be only accessible behind CMS VPN. In addition to locking down the environments to be accessible behind VPN changes were required to temporarily enable access for GitHub Actions to access the environment to run tests.

Because the serverless environment uses AWS Cloudfront we cannot place Cloudfront inside of a private subnet. Therefore, to restrict traffic the AWS WAF is being utilized to lock down access. To accomplish this we're beginning to utilize the [CMS WAF Serverless plugin](https://github.com/Enterprise-CMCS/serverless-waf-plugin) to apply standard rule sets and extend that by applying custom IP sets.

This pull request does the following:
1) Implements the CMS WAF Plugin 
2) Creates a prioritized rule set for the WAF 
3) Creates a custom IP set per stack 
4) Associates a manually created global IP set that houses all CMS VPN IP addresses to the stacks WAF
5) Looks up GitHub Runner IP's and adds them to the stacks custom IP set (that is only applied to that stack) 
6) Upon successful run of all tests clears out the GitHub Runner IP's from the stacks custom IP set

### Related ticket(s)

https://jiraent.cms.gov/browse/CMDCT-2942

---
### How to test
To test locally connect to Zscaler or Cisco Any Connect:
In an incognito window navigate to the endpoint of the environment and you should see the login page for QMR and should be able to login and use the app as expected 
Disconnect from Zscaler and navigate to the same endpoint in a new incognito window and you should receive a 403. 

### Important updates
When creating new projects or migrating this work to other products we will need to manually create the CMS VPN IP set

**To implement this in another product the following change also needs to be done for merging to existing environments** 
1) manually create a Global (cloudfront) waf ip set named "cmsvpn" in both the dev and val aws accounts populating CIDR blocks for cisco any connect as well as Zscaler. (Work with team leads for those IP's if needed)
2) Create ssm parameter for the IP set container the ARN 
Follow the changes below as well:
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2037
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2045

### Additional Details
The script that registers IPs against the IP set assigned to the WAF for the application stack, is designed to be able to be run concurrently while managing race and optimistic concurrency control locks.  The new implementation of the web application firewall (version 2) requires setting the entirety of the array in each operation, and does not support append.  To avoid concurrency write contention, AWS requires a lock token be supplied in the update operation, such that a read must occur before a write.  Console actions will also invalidate a token.  Read operations are also throttled, so this script implements an exponential backoff strategy with jitter, such that a random time between min, and the lesser of min^try or max, is used to sleep between each each read, each write, and the entirety of the loop, with short circuits should the number of attempts of any operation exceed the circuit breaker number of attempts.  This allows for a relatively high level of concurrency, at the cost of performance with creating the registry.  For other implementation patterns, the not fully realized branch, `skipci-whitelist-gha-cidrs` can be used an example.

Also worth noting the reason that a global vpn ip set and a per stack GitHub runner ip set was chosen was to allow a single place with ideally few updates to control VPN access. Then, for the individual stacks IP set we can simply add the GitHub runner ip's, run tests, then clear out those IP's at the end of the run ensuring that we only apply this to one stack and ensure security across all the stacks. 

The script is called as follows:
`./waf-controller.sh [set|append] [ip_set name] [ip_set id] [space delimited list of ips or CIDRs]`

---
### Author checklist

---

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -76,11 +76,11 @@ custom:
             SampledRequestsEnabled: true
             CloudWatchMetricsEnabled: true
             MetricName: ${self:service}-${self:custom.stage}-block-traffic
-            Statement:
-              NotStatement:
-                Statement:
-                  IPSetReferenceStatement:
-                    ARN: ${self:custom.vpnIpSetArn}
+          Statement:
+            NotStatement:
+              Statement:
+                IPSetReferenceStatement:
+                  ARN: ${self:custom.vpnIpSetArn}
   firehoseStreamName: aws-waf-logs-${self:service}-${self:custom.stage}-firehose
   scripts:
     hooks:


services/ui/serverless.yml
```diff
@@ -137,35 +199,6 @@ resources:
                 Bool:
                   aws:SecureTransport: false
         Bucket: !Ref CloudFrontLoggingBucket
-    CloudFrontWebAcl:
```
braxex, line 140: Question: Do we still need these rules applied in some way for val and prod?

braxex (approved):
Looks good to me. Left one question.

BearHanded (commented):
I'm not seeing the 403 on access when I leave the vpn, am I missing something?

BearHanded (approved):
looks like it handles the 403 now, nice

services/ui/serverless.yml
braxex: Ah nice. I figured there was something in that plugin but I poked around a bit and didn't immediately see it so I figured I'd just ask. Works for me!

braxex (approved):

PR #2046, opened 2024-02-08
Fix infrastructure destroy actions

berryd:
### Description

The purpose of this pull request is to correct destroy actions for Snyk or Dependabot created branches by providing a consistent means to generate shortened stage names from autogenerated branches that may exceed Cloudformation length requirements when prefixed to services managed by Cloudformation stacks.  Additionally, this pull request adds the ability to trigger destroy actions manually against any non-protected target using any Github hosted branch.

- `GITHUB_REF` environment variable is the same as `github.ref` and includes the fully qualified path
- Destroy actions always trigger on the default branch so `GITHUB_REF`/`github.ref` would be `/refs/heads/master`
- Destroy actions know what the target is based on `github.event.ref` which is NOT the fully qualified path
- Defining `workflow_dispatch` enables manual triggering destroy actions against a given code base and enables specifying target.

The changes in this pull request do the following:
- Apply consistent lookups in Deploy (concurrency lock would be deploy-[branch] rather than deploy-/refs/heads/[branch])
- Remove hard coded addition to destroy action which would break if a tag is deleted
- Enable manual dispatching of events.

For other repositories which may differ in structure, it is suggested that the `workflow_dispatch` be added as an initial pull request, as this has to make it to the default stage before it can be acted upon.  This will enable development against destroy actions which can only trigger from the default branch until the configuration change is merged.

### Related ticket(s)

CMDCT-3257

---
### How to test

Once this has been merged...

Create three branches

dependabot/npm_and_yarn/berry-did-this
snyk-upgrade-berry-did-this9faf06c882b46cc4a8e83
berry-test
Push and wait for builds to succeed.

Delete the remote branches and verify the destroy step completes.

Create one more branch, any arbitrary name, and add echo this is a test && exit 1 to the destroy yaml file at line 37. Exercise the manual trigger and verify the update in the unmerged branch is executed in the destroy step

### Important updates

No functional changes should be experienced during normal development activities.  The changes do include some minor refactoring of how references are leveraged, but the functional changes should apply to Snyk and Dependabot created branches with the addition of Github actions manual triggering from the destroy workflow.

As of the time of writing of this pull request, the following are the remote branches for this product, and the highlighted ones are the ones this pull requests asserts to be protected:
  origin/berry-test
  origin/deployed-login
  origin/entjira
  origin/fix-snykdependa-destroy
  origin/iam
  **origin/master**
  **origin/prod**
  origin/skipci-fix-snyc-jira-sync
  origin/skipci-whitelist-gha-cidrs
  origin/snyk-fix-2b8efea0f3610daa3da7981342c7806d
  origin/snyk-upgrade-0423a0cea1480868aaa764d684553e91
  origin/snyk-upgrade-051019ec32e59ecdb41dc5984b09a5b2
  origin/snyk-upgrade-796c34068a569533c3116c7807574d9c
  origin/snyk-upgrade-8aa0ba3488495e9c428d253b5eada593
  origin/snyk-upgrade-ba6915d13282c86e0c0f201e6a32b181
  origin/snyk-upgrade-ce82709dd58d0031edf2aa9149be08e2
  origin/snyk-upgrade-e23107549426666253c1ebf267ca2955
  origin/tealium-3288
  **origin/val**

---
### Author checklist

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -27,7 +27,7 @@ jobs:
     env:
       SLS_DEPRECATION_DISABLE: "*" # Turn off deprecation warnings in the pipeline
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: set branch_name # Some integrations (Dependabot & Snyk) build very long branch names. This is a switch to make long branch names shorter.
         run: |
           BRANCH_NAME=$(./.github/setBranchName.sh ${{ github.ref_name }})
@@ -99,7 +99,7 @@ jobs:
         run: |
           echo "No endpoint set, Check if the deploy workflow was successful."
           exit 1
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: set branch_name
         run: |
           BRANCH_NAME=$(./.github/setBranchName.sh ${{ github.ref_name }})
@@ -315,7 +315,7 @@ jobs:
     env:
       SLS_DEPRECATION_DISABLE: "*" # Turn off deprecation warnings in the pipeline
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Configure AWS credentials for GitHub Actions
         uses: aws-actions/configure-aws-credentials@v4
         with:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ on:
       - "!skipci*"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
 
 permissions:
   id-token: write
@@ -19,7 +19,7 @@ jobs:
   unit-tests:
     name: Unit Tests
     uses: ./.github/workflows/unittest-workflow.yml
-    if: github.ref == 'refs/heads/master'
+    if: github.ref_name == 'master'
     secrets:
       CODE_CLIMATE_ID: ${{ secrets.CODE_CLIMATE_ID }}
   deploy:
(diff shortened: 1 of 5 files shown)


.github/workflows/destroy.yml
```diff
@@ -14,17 +20,23 @@ jobs:
...
-    if: github.event.ref_type == 'branch' && !contains(fromJson('["master", "val", "prod"]'), github.event.ref)
+    if: |
+      (
+        github.event.ref_type == 'branch' &&
+        (!startsWith(github.event.ref, 'skipci')) &&
+        (!contains(fromJson('["master", "val", "production"]'), github.event.ref))
+      ) ||
+      (
+        inputs.environment != '' &&
+        (!contains(fromJson('["master", "val", "production"]'), inputs.environment))
```
gmrabian, line 31: `prod` in QMR, not `production`

.github/workflows/destroy.yml
```diff
@@ -14,17 +20,23 @@ jobs:
...
+      )
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
```
gmrabian, line 35: checkout@v4 is out if we want to upgrade while we're here 🤷 not really relevant though

.github/workflows/deploy.yml
```diff
@@ -19,25 +19,19 @@ jobs:
...
-    if: github.ref == 'refs/heads/master'
+    if: github.ref_name == 'master'
     secrets:
       CODE_CLIMATE_ID: ${{ secrets.CODE_CLIMATE_ID }}
   deploy:
     runs-on: ubuntu-latest
     env:
       SLS_DEPRECATION_DISABLE: "*" # Turn off deprecation warnings in the pipeline
     steps:
+      - uses: actions/checkout@v3
```
gmrabian, line 30: checkout@v4 is out if we want to upgrade while we're here 🤷 not really relevant though

gmrabian (changes requested):
branch name change but otherwise looks ok to me

.github/workflows/destroy.yml
berryd: Good catch

.github/workflows/destroy.yml
berryd: I'm good with doing this while I'm in here.  Changes made and pushed.

PR #2053, opened 2024-02-13
CMDCT-3295: Landing Page Content Update

karla-vm:
### Description

For the FFY2024 landing page, the BOs have requested the removal of this text:

Screenshot 2024-02-13 at 3 21 35 PM

### Related ticket(s)

CMDCT-3295

---
### How to test

- Log in as a state user

You should not see that text area for 2024, but it should be visible for other years.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/views/StateHome/AddCoreSetCards.tsx
+++ b/services/ui-src/src/views/StateHome/AddCoreSetCards.tsx
@@ -82,7 +82,7 @@ export const AddCoreSetCards = ({
           coreSetExists={healthHomesCoreSetExists}
         />
       )}
-      {year !== "2024" && (
+      {year && parseInt(year) >= 2024 && (
         <CUI.Center w="44" textAlign="center">
           <CUI.Text fontStyle="italic" fontSize="sm">
             Only one group of Adult Core Set Measures can be submitted per FFY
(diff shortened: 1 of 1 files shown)


services/ui-src/src/views/StateHome/AddCoreSetCards.tsx
```diff
@@ -81,11 +82,13 @@ export const AddCoreSetCards = ({
           coreSetExists={healthHomesCoreSetExists}
         />
       )}
-      <CUI.Center w="44" textAlign="center">
-        <CUI.Text fontStyle="italic" fontSize="sm">
-          Only one group of Adult Core Set Measures can be submitted per FFY
-        </CUI.Text>
-      </CUI.Center>
+      {year !== "2024" && (
```
BearHanded, line 85: This likely should reflect years going forward so we don't forget next year, or some centralized location that tracks years vs features, if that's a pattern we're falling into. I know some things are already tracking 2023+

BearHanded (changes requested):

services/ui-src/src/views/StateHome/AddCoreSetCards.tsx
karla-vm: Good point, let me ask David if this is going to be removed going forward or ONLY for 2024.

BearHanded (approved):

PR #2055, opened 2024-02-15
[QMR Cleanup] Removing  types.ts From Each Measures P1

ailZhou:
### Description

**Summary**
This is going to be one of many PRs related to cleaning up QMR and reducing unit test / commit runtime to something a little more reasonable. My goal is for each PR to have incremental changes, despite how many files are change, it shouldn't be a large change in structure at one go.

**The Update**
Do not be scared off by the sheer amount of files changed here, it was the same code update repeated hundreds of times. In each measure's folder there was a types.ts file. In that file was usually this chunk of code:

Screenshot 2024-02-15 at 4 28 09 PM

What I did was remove the file & changed the import from:
Screenshot 2024-02-15 at 4 30 48 PM

to 

Screenshot 2024-02-15 at 4 30 53 PM

And that change only affected index.tsx & validation.ts  or just validation.ts

**NOTE:** A total of 4 measure's types.ts deviated from this pattern, and will have their own PR to address their types.ts file:
- AMRAD
- CPAAD
- CPCCH
- MSCAD

**Why change it?**
The types.ts files were being repeated hundreds of times and  (based on my best guess) all it did was centralize where the type information is for the index & validation file. It would have been a good idea if QMR wasn't so massive but with us having to copy everything year over year, these just created a lot of redundant files that did very little and created too much repetition.

### Related ticket(s)

CMDCT-3300

---
### How to test

If you can run QMR and save normally, that is enough.

### Important updates

N/A

---
### Author checklist

---


BearHanded (approved):
Confirmed it runs locally.

gmrabian (approved):
ran locally; no console errors when clicking around

PR #2057, opened 2024-02-16
Fix for ipv6

dwhitestratiform:
### Description
In debugging with the cms networking folks we have discovered that we need to add ipv6 support (specifically for Zscaler) to our vpn restricted environments. In the following ticket the networking folks gave ipv6 cidr blocks that were added to a separate ipv6 ip set. Note: in aws ip sets can only be ipv4 or ipv6 therefore we need to create a unique ip set for each. Also, a change was added to use an "or" rule where any of the following can be true 
if on cms vpn with ipv4 ip 
if on cms vpn with ipv6 ip 
if an ip in the GitHub ip set 
allow traffic
else 
block all traffic 

be advised ... the Zscaler team has let us know that ipv6 is currently not enabled on their end however, when they do turn on that support this will 'just work' . The ETA on ipv6 support for Zscaler is mid to late march 2024

### Related ticket(s)

---
### How to test
ensure you can get to the endpoint of the environment built from this branch when on Zscaler

### Important updates
this requires a new ip set be created by hand for each product and a new arn be added to ssm to be pulled at deploy time.

---
### Author checklist

- [x I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
---

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v2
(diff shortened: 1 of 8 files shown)


gmrabian (changes requested):
I was not able to reach the deployed URL on or off zscaler. I tried two browsers and private/incognito mode
tested with:
https://dq28xlp8wc56i.cloudfront.net/

gmrabian (approved):

PR #2064, opened 2024-02-23
Refactor: Additional Notes

ailZhou:
### Description

First CommonQuestions to be refactored: Additional Notes.

A lot of the components in the commonQuestions folder are very same/similar between years, usually the labels are different but the functionality tends to be the same. In a means to centralize these components to a shared folder for all the years, the labels will be striped out to their own individual year's data object. 

Screenshot 2024-02-26 at 3 43 22 PM

I moved the AdditionalNotes component to its own folder to be shared between all four years

Screenshot 2024-02-26 at 3 44 54 PM

I created a new react context called "SharedContext", to load this information instead of passing it as a props due to the sheer amount of unit test that would need to be updated. Once more scripts have been refactored, I will update it to being passed through a prop.

Screenshot 2024-02-26 at 3 46 07 PM

There are a couple of instances where the comments of the code have the words "WIP", this is for tracking purposes. It means this chunk of code is just to be used during the factor process to keep PRs readable and be used as scaffolding until more pieces have been disentangled.  

Screenshot 2024-02-26 at 3 48 03 PM

- This piece of code is set up this way due to the unit & cypress test not responding well to dynamic imports atm. Once there is less duplicate test, I will switch over.

### Related ticket(s)

CMDCT-3300

---
### How to test

QMR runs and Additional Notes labels look the same as they did before.

Link to val for ref:
https://d1bf3evxyzlx5y.cloudfront.net/

### Important updates

---
### Author checklist

---


BearHanded (dismissed):

services/ui-src/src/labels/2023/commonQuestionsLabel.tsx
```diff
@@ -0,0 +1,11 @@
+export const commonQuestionsLabel = {
+  AdditonalNotes: {
+    header: "Additional Notes/Comments on the measure (optional)",
+    section:
+      "Please add any additional notes or comments on the measure not otherwise captured above (<em>text in this field is included in publicly-reported state-specific comments</em>):",
```
ajaitasaini, line 5: does 2021 and 2022 not need the extra text in the <em> tags?

services/ui-src/src/shared/types/Type.tsx
```diff
@@ -0,0 +1 @@
+export * from "shared/types/TypeAdditionalNotes";
```
keeysnc, line 1: Suggestion: Since were in the types directory could this file just be called index.tsx?

services/ui-src/src/utils/parser.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { v4 as uuidv4 } from "uuid";
+
+//convert labels with html tags to JSX elements to be rendered
+export const parseLabelToHTML = (label: string) => {
+  const parseLabel = new DOMParser().parseFromString(label, "text/html").body;
+
+  let renderArr: (
+    | string
```
keeysnc, line 9: I think the first `|` in the renderArr can be removed here. (line 9)

keeysnc (commented):
Everything looks good, I just had a few comments and suggestions.

services/ui-src/src/shared/types/Type.tsx
ailZhou: we could but I think that might cause more confusion than is helpful.

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -142,6 +144,10 @@ export const MeasureWrapper = ({
     []
   );
 
+  //WIP: this code will be replaced with a dynamic import onces we refactored enough files
+  const shared =
```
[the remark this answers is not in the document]
ailZhou: this a temporary scaffolding code so its not necessary to add a type to it. It would just be an AnyObject

services/ui-src/src/utils/parser.tsx
keeysnc: Ah prettier okay yea if it didn't cause any issues it should be fine.

keeysnc (dismissed):

keeysnc (approved):

services/ui-src/src/utils/parser.tsx
benmartin-coforma, line 1: This doesn't seem right to me.
1. It really seems like `for..of` or `.map` would be a better choice than `.forEach` - although when I tried that I found that TS was very unhappy about it, and I couldn't manage to fix that.
    * Personally I'd still write a helper function that converts `NodeListOf<T extends Node>` into `T[]`, but you might decide it's cleaner not to.
    * One advantage of using `.map` to build the array is that you don't have to specify the type up front, which reduces the noise a bit.
    * Code below.
2. It seems that checks for `node instanceof Text` and `node instanceof Element` would be neater than checks for the truthiness of `nodeValue`
    * No inference would be required on the part of the developer
    * No casts would be required in the code
    * It also makes for a natural `else { throw new Error("Unrecognized node type in label:\n" + label) }` clause, to warn any future devs that attempt to embed Comments or CDATA or any other odd thing in the future.
3. This method doesn't recurse, so any markup more complex than a single tag will get mangled - for example, an `<a>` tag inside of a `<p>` would be wiped out, and we would only see the link text.
    * Failing test below.
    
```ts
/**
 * This helper function shouldn't be necessary,
 * but TS insists NodeListOf<T> does not have Symbol.iterator.
 * If you can fix that, please delete this function.
 */
function arrayFromNodeList <T extends Node>(nodeList: NodeListOf<T>) {
  const arr: T[] = [];
  nodeList.forEach((node) => arr.push(node));
  return arr;
}
```

```ts
  it("Test that it renders tags recursively", () => {
    const label = "<em><strong>hello</strong></em>";

    const parsedLabel = parseLabelToHTML(label);
    const { container } = render(parsedLabel);

    const outerTag = container.querySelector("em");
    expect(outerTag).toBeTruthy();
    const innerTag = container.querySelector("strong");
    expect(innerTag).toBeTruthy();
  });
```

benmartin-coforma (approved):
Approved, with the understanding that label parsing will be revisited soon.

PR #2065, opened 2024-02-26
Purify incoming html body

gmrabian:
### Description

Resolves a security concern that a malicious user can intercept the network call containing base64 encoded HTML and insert dangerous iframes, scripts, etc that act on the downstream system. Now any received HTML is sanitized before sending downstream.

We use DOMPurify in MCR and MFP to [sanitize all api requests](https://github.com/Enterprise-CMCS/macpro-mdct-mcr/blob/main/services/app-api/handlers/handler-lib.ts#L31). May be worth considering for the future in other apps.

references:
[dompurify](https://github.com/cure53/DOMPurify)

### Related ticket(s)

CMDCT-3330

---
### How to test

Tool for encoding/decoding base64 strings
https://examples.aswsec.com/tools/encode-decode.html#

- Add a `console.log("sanitizedBody", sanitizedBody);` on line 23 of pdf.ts
- Start QMR locally
- Log in as a state user
- On the Homepage click on the three vertical dots on the adult measures row
- Select "Export"
- Open the browser tools' network tab
- Click "Print PDF" in the UI (the request will fail because it cannot connect to prince)
- Modify the network request to use this base64 string
`PHA+YWJjPGlmcmFtZS8vc3JjPWpBdmEmVGFiO3NjcmlwdDphbGVydCgzKT5kZWY8L3A+`
This decodes to `<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>`, which contains a potentially dangerous iframe
- Ensure the log in your terminal shows `<p>abc</p>` as the sanitized output 🎉 

---
### Author checklist

---

### DIFF
--- a/services/app-api/handlers/prince/pdf.ts
+++ b/services/app-api/handlers/prince/pdf.ts
@@ -55,7 +55,7 @@ export const getPDF = handler(async (event, _context) => {
     headers: {
       host: hostname, // Prince requires this to be signed
     },
-    encodedSanitizedBody,
+    body: encodedSanitizedBody,
   };
 
   const signer = new SignatureV4({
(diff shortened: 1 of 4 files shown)


berryd (approved):

ailZhou (approved):

BearHanded (commented):
tested and verified

PR #2076, opened 2024-02-29
[Fix] Parser to be Recursive

ailZhou:
### Description

Upgrading the html parser based on @benmartin-coforma feedback.
1) Change conditionals to check type of node instead of truthiness and added the error recommended.
2) Parser now recurse through the string for any inner tags. Nested tags should not be a problem anymore.
3) Added the recursive test that Ben provided into the unit test and also added a new one to test the error message.

Note: I choose to stay with the forEach loop instead of converting it to an array to be able to use map.

### Related ticket(s)

CMDCT-3338

---
### How to test

1) Sign into QMR
2) Stay in reporting year 2024
3) Go to any adult measures
4) Addtional Note should render like this:

> Please add any additional notes or comments on the measure not otherwise captured above (_text in this field is included in publicly-reported state-specific comments_):

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/parser.test.ts
+++ b/services/ui-src/src/utils/parser.test.ts
@@ -26,4 +26,12 @@ describe("Test parseLabelToHTML", () => {
       "Unrecognized node type in label:\n" + label
     );
   });
+
+  it("Test that attributes are copied", () => {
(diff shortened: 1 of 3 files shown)


benmartin-coforma (approved):
As co-author, I approve.

PR #2079, opened 2024-02-29
Implement DocRaptor for PDF generation

benmartin-coforma:
### Description
Instead of sending HTML to CMS's Prince instance to be rendered to PDF, send it to DocRaptor's instance.

### Related ticket(s)
CMDCT-3249

---
### How to test
* Attempt to print a PDF
   * There are two ways to do this in QMR. The big "print" button at the top of each measure page uses browser print, but the "export" option for an entire core set uses the API print. Don't do the browser print one.
* You should see a PDF
   * If you don't, check your browser's address bar to see if it blocked a pop-up. That'll be it.
* The watermark on that PDF is expected
   * Even though I didn't see one when I tested in this branch's environment? Maybe something to do with the DR pricing tier CMS selected.
   * But I verified in DR's logs that the `test: true` parameter definitely came through, and it was counted as a test request.

### Important updates
There is a new environment variable, `docraptorApiKey`. You only need to provide a value if you're trying to test PDF generation locally. Either sign up for a free trial key at DocRaptor.com, or ask me for my key - I'm happy to share it privately.

---
### Author checklist

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -23,3 +23,5 @@ MEASURE_TABLE_NAME=local-measures
 # LAUNCHDARKLY
 LD_PROJECT_KEY=mdct-qmr
 LD_SDK_KEY=sdk-25b0f45f-bb20-4223-aede-32d66b525721 #pragma allowlist secret
+
+docraptorApiKey=YOUR_API_KEY_HERE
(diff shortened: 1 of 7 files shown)


ailZhou (approved):
lgtm! was able to get a response pdf, no watermark though.

PR #2080, opened 2024-03-01
Upgrade to AWS SDK v3

benmartin-coforma:
### Description
Upgrade away from AWS SDKv2, as much as practical.

This involved sweeping changes to the API, and minimal changes elsewhere. Our code in the uploads service still relies on SDKv2, because of conflicts with the node version and the stream functionality in the SDKv3 S3 library. And all of our services still indirectly rely on SDKv2, because Serverless still uses it.

One notable decision I made was to modify the interface of dynamodb-lib. Previously, it returned result objects shaped like `{ Count: number, Items: T[] }`. Now, it returns the `T[]` directly. This makes it more convenient to consume in the API handlers. However, that result object structure was also return directly from the API handlers to the UI - there are many many places in the UI where we unwrap a result with `data?.Items`. I believe these should all also be changed, but elected not to do so at this time; this PR was already large enough. That's why, in some cases, the API handlers now take the array from Dynamo and _re-wrap_ it in .Items, before returning it to the UI.

### Related ticket(s)
CMDCT-3075

---
### How to test
* Root: Removed a dependency. Test by verifying that `dev local` still works.
* App-api:
  * Banners: Banners/fetch.ts was changed. Confirm that banners display properly after being created.
  * Core Sets: Confirm that core sets can be created, updated, and deleted.
    * Additionally, confirm that the numComplete and numAvailable are correctly set on fetch.
  * Measures: Confirm that measures can be created, updated, and deleted.
    * Additionally, confirm that autocompleted measures are correctly marked as such.
  * Debug-lib: Confirm that lambda logs in CloudWatch contain an initial log from the handler, as well as logs from within AWS's Dynamo library.
  * Kafka: Confirm that topics are written to. Note that we expect an error from the success logging message on line 130 of kafka-source-lib; that fix is out of scope for this PR.
* Database: Updated some migrations. No need to test.
* Stream-functions: Minor updates here, made irrelevant by the deletion of this service since then
* Ui-auth: Updated user creation code. This is actually hard to test, because all dev logins share the `master` cognito pool. Check the ephemeral cognito pool for this branch to verify users exist there.
* Ui-src: Deleted unused library, and comments referring to it. It was intended to be used for local S3 work, but I suspect it never was.
* Uploads: No changes here; our hands are tied by the node version. A follow-up ticket will be created to upgrade this service.

### Important updates
Yep, you'll have to `yarn` in most directories after pulling this.

---
### Author checklist


services/app-api/handlers/banners/tests/proxyEvent.ts
```diff
@@ -10,60 +10,17 @@ export const testBanner = {
...
-      userArn: "",
-      clientCert: {
-        clientCertPem: "",
-        issuerDN: "",
-        serialNumber: "",
-        subjectDN: "",
-        validity: { notAfter: "", notBefore: "" },
-      },
-    },
+    /* nope */
```
BearHanded, line 24: 👨‍🍳 👌

services/stream-functions/libs/ses-lib.js
BearHanded, line 1: Depending on the timing of this PR vs the remove email handler ticket this file might just be gone

benmartin-coforma: Right you are. I merged master and resolved the conflict; this file is gone now.

ajaitasaini (approved):
tested together!

BearHanded (approved):

PR #2082, opened 2024-03-01
CMDCT-3324: ADD-CH Content Updates (FFY2024)

karla-vm:
### Description

Updating the `Performance Measure` description section of **ADD-CH** to include numbering as seen below:

Screenshot 2024-03-01 at 11 12 18 AM

### Related ticket(s)

CMDCT-3324

---
### How to test

- Log in as a state user
- Add child core sets
- Navigate to ADD-CH

Verify that the `Performance Measures` description matches the screenshot.

### Important updates

---
### Author checklist

---


ailZhou (changes requested):
questionSubtext is also used by AIF-HH so its turning AIF-HH text into using numbers list as well. I recommend following the pattern of "questionListItems"  and creating a new "questionListOrderedItems". That way we can reuse it for CCP-CH

karla-vm:
> questionSubtext is also used by AIF-HH so its turning AIF-HH text into using numbers list as well. I recommend following the pattern of "questionListItems" and creating a new "questionListOrderedItems". That way we can reuse it for CCP-CH

great call, I made those changes! thank you ✨

ailZhou (approved):
LGTM! Thank you for making the changes!

PR #2139, opened 2024-03-22
Run a11y tests in parallel to speed up pipeline

dwhitestratiform:
### Description
This PR is to run the a11y accessibility tests in parallel to the e2e tests in an effort to cut down the deploy time of QMR. 

Experimenting it looks like a previous run where the a11y tests run after the e2e tests on a second run (not initial deploy) takes about 19.5 mins 
Screenshot 2024-03-21 at 8 22 19 PM

Moving the a11y tests to run in parallel and not after e2e is ~3.5 mins shorter and runs through without issue.
Screenshot 2024-03-21 at 8 17 30 PM

### Related ticket(s)
https://jiraent.cms.gov/browse/CMDCT-3438
---
### How to test
keep an eye on the pipeline an ensure we don't see adverse effects of running this in parallel but initial testing looks good. 

### Important updates
na/

---
### Author checklist

---

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -314,7 +314,6 @@ jobs:
     name: Accessibility Tests
     needs:
       - e2e-tests-init
-      # - setup-tests
     uses: ./.github/workflows/cypress-workflow.yml
     with:
       test-path: "a11y"


ailZhou (approved):
ii'll approve anything that makes these test run faster

PR #2142, opened 2024-03-22
Change styles and DOMPurify config to make PDFs look better

benmartin-coforma:
### Description
We apply DOMPurify to print requests to avoid forwarding potentially-dangerous HTML to DocRaptor. But DOMPurify strips out rather a lot from our HTML. This PR puts some of it back.

1. DOMPurify, by default, returns only the sanitized inner HTML of the `<body>`. The outer `<html>` is restored by passing the `WHOLE_DOCUMENT` option, and the `<head>` is restored with `ADD_TAGS: ["head"]`.
2. DOMPurify removes any `<style>` tag that contains a `<` character. This is a problem because we include styles with example HTML, embedded as comments within the CSS. This is addressed by stripping out CSS comments on the browser side, before making the call to our API's print endpoint.
   * This removes a few dozen CSS comments. Most are linter directives, and a few are explanations or examples. Nothing is lost here. However we also have CSS source maps embedded as comments within `<style>` tags, so in theory our app's styles become more difficult for developers to debug after printing a PDF. These source maps are restored on page refresh, so this is not a huge loss either; I'm just noting it here.
3. DOMPurify, by default, removes all `<link>` tags. We use links to include the Open Sans font, as well as Bootstrap. I am of the opinion that DOMPurify is using an overabundance of caution here - I personally doubt that allowing `<link>` tags to be sent to DocRaptor poses a significant security risk to them, and I am personally confident that it poses no security risk to us. So I restore them here with `ADD_TAGS: ["link"]`.
4. I also tweaked the endpoint's input validation, adding a check and a new error message for the sake of a pre-existing unit test. I don't expect this new check or message to give us much value in practice, so I would be open to removing it (and the corresponding test). But I don't mind including it either; there's an outside chance it may help a future developer debug a PDF issue someday.
5. DocRaptor requires that all resources on the page are loaded from absolute URLs (because relative URLs appear to them as filesystem access attempts). Rather than tracking down & transforming all relative URLs on the page, I added a `<base>` tag pointing to the current domain. This also required me to `ADD_TAGS: ["base"]`, to prevent DOMPurify from wiping it out.
6. This does not represent a complete "fix" for all styling issues within the PDF. For example, I believe the color of the measure buttons at the very top should be a light gray, and I'm not seeing that. But it's a couple of big steps forward, and I would prefer to prioritize troubleshooting any minor remaining issues separately from this ticket.
   * I suspect that the remaining visual issues in the PDF are due to errors on DocRaptor's side, when attempting to load our resources. There may also be some inherent issues in our CSS, which our browsers might be handling differently from DR's renderer. That is what I gather from the logs viewable on DocRaptor's website; I may not be reading them correctly.

---
Future developers troubleshooting DOMPurify-related issues: The approach I found most effective for tracking down problems was to modify the print endpoint to write the HTML to file, just before sanitization. Then I wrote a script that looked like this:
```js
const fs = require("fs");
const createDOMPurify = require("dompurify");
const { JSDOM } = require("jsdom");
const windowEmulator = new JSDOM("").window;
const DOMPurify = createDOMPurify(windowEmulator);
const rawHtml = fs.readFileSync("before.html");
const sanitizedHtml = DOMPurify.sanitize(rawHtml, {
    WHOLE_DOCUMENT: true,
    ADD_TAGS: ["head", "link"],
  });
fs.writeFileSync("after.html", sanitizedHtml);
```
After which, I repeatedly made manual modifications to `before.html`, and invoked 
```bash
node test.js && my-favorite-difftool before.html after.html
```
This workflow let me see exactly what DOMPurify was doing, and to quickly test out sweeping changes (like deleting 1000 lines of CSS) in order to narrow down what could be triggering DOMPurify's behavior.

I will also note that the resulting diff will be _very_ noisy, due to DOMPurify's tendency to rearrange attributes. For example `<foo bar="1" baz="2"/>` might become `<foo baz="2" bar="1/>`. While this has no effect on the rendering of the page, it does greatly confuse any text-based diffing algorithm.

---
### Related ticket(s)
CMDCT-3370

---
### How to test
1. Log in to QMR
2. In the 3-dot menu for a core set, choose Export
3. Click Print at the top of the page
4. Look at the PDF. It should appear generally correct.

### Important updates

---
### Author checklist

### DIFF
--- a/services/app-api/handlers/prince/pdf.ts
+++ b/services/app-api/handlers/prince/pdf.ts
@@ -12,39 +12,20 @@ export const getPDF = handler(async (event, _context) => {
   if (!rawBody) {
     throw new Error("Missing request body");
   }
-
   if (rawBody.startsWith("{")) {
     throw new Error("Body must be base64-encoded HTML, not a JSON object");
   }
-
-  if (!DOMPurify.isSupported) {
-    throw new Error("Could not process request");
-  }
-
-  // decode body from base64, sanitize dangerous html
-  const decodedBody = Buffer.from(rawBody, "base64").toString();
-  const sanitizedBody = DOMPurify.sanitize(decodedBody, {
-    WHOLE_DOCUMENT: true,
-    /*
-     * By default DOMPurify removes all <link> tags, due to the possibility
-     * of XSS in certain browsers. We include bootstrap styles and google
-     * fonts via <link> tags in QMR, and we don't expect this page to be
-     * rendered in vulnerable browsers. So we allowlist "link" here.
-     *
-     * https://github.com/cure53/DOMPurify/issues/2
-     */
-    ADD_TAGS: ["head", "link"],
-  });
-
   const { docraptorApiKey, stage } = process.env;
   if (!docraptorApiKey) {
     throw new Error("No config found to make request to PDF API");
   }
 
+  const decodedHtml = Buffer.from(rawBody, "base64").toString();
+  const sanitizedHtml = sanitizeHtml(decodedHtml);
   const requestBody = {
     user_credentials: docraptorApiKey,
     doc: {
-      document_content: sanitizedBody,
+      document_content: sanitizedHtml,
       type: "pdf" as const,
       // This tag differentiates QMR and CARTS requests in DocRaptor's logs.
       tag: "QMR",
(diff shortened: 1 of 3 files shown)


BearHanded:
Is this what I should be seeing?
image

Looks like more is getting removed than expected?

BearHanded (approved):
tested on the deployed branch, looks better.

ailZhou (approved):
tested in deploy, saw pdf. missing some styling but still readable

PR #2146, opened 2024-03-25
Remove the rateReadOnly prop from PerformanceMeasure; it is unused

benmartin-coforma:
### Description
Remove the `rateReadOnly` prop from the 2024 PerformanceMeasure component; it is unused.

The only component that could possibly have been affected by this was MSC-AD, which explicitly passed in `false`, overriding the possibility of becoming readonly due to the data source. But, (1) during my testing I did not see any change between passing `false` and passing nothing; and (2) MSC-AD no longer passes `false` anyway, now that it is an autocomplete measure.

Rates can still become readonly, depending on the datasource and perhaps other possible factors, but never via this prop.

### Related ticket(s)
N/A

---
### How to test
As dead code removal, this PR can only be regression tested; Cypress will do a decent job of that.

### Important updates
N/A

---
### Author checklist


BearHanded (approved):
approved assuming unit tests pass, rerunning them. It weirdly looks like the action for codeclimate failed

PR #2147, opened 2024-03-25
Consolidate DataSourceCahps Questions

keeysnc:
### Description
Consolidation of individual year DataSourceCahps files and rendering correct labels based on the year.

### Related ticket(s)
CMDCT-https://jiraent.cms.gov/browse/CMDCT-3443

---
### How to test
1. Select each year and select a measure that shows a Data Source section.  ( ex. AMM-AD & CCS-AD)

Screenshot 2024-03-25 at 3 07 40 PM

2. Fill out the measure and select "Other Data Source" when you get to the "Data Source" section. The following labels should show up:

* Years 2021 and 2022 should have "Describe the data source:"
* Years 2023 and 2024 should have "Describe the data source (text in this field is included in publicly-reported state-specific comments):"

### Important updates

---
### Author checklist

---


services/ui-src/src/shared/commonQuestions/DataSourceCahps.tsx
```diff
@@ -30,7 +35,7 @@ export const DataSourceRadio = ({ data = defaultData }: DataSourceProps) => {
             children: [
               <QMR.TextArea
                 name="describeTheDataSource"
-                label="Describe the data source:"
+                label={parseLabelToHTML(labels?.DataSource.describeDataSrc)}
```
ailZhou, line 38: Request: For ease of future update, we don't want to have the components reuse other components labels as this can cause unintended text updates. Can you add a new key `DataSourceCahps:{describeDataSrc: ""}`  into each years of the commonQuestionsLabel.tsx and switch the reference over so that it has its own label

ailZhou (approved):
lgtm! tested in browser and text matches the year

side note: testing should be used for FVA-AD as that's the one this is used for.

PR #2162, opened 2024-04-04
CMDCT-3478: Content Removal (Global)

karla-vm:
### Description

For FFY2024 onwards, removing the following text box from `Definition of Population Included in Measure` section of a measure:

image

### Related ticket(s)

CMDCT-3478

---
### How to test

- Log in as a state user
- Navigate to any measure in 2024
- Verify that the text box is not visible

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
+++ b/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
@@ -11,7 +11,6 @@ import * as DC from "dataConstants";
 import { useContext } from "react";
 import SharedContext from "shared/SharedContext";
 import { AnyObject } from "types";
-import { useParams } from "react-router-dom";
 
 interface Props {
   childMeasure?: boolean;
@@ -169,7 +168,6 @@ export const DefinitionOfPopulation = ({
   healthHomeMeasure,
 }: Props) => {
   const register = useCustomRegister<Types.DefinitionOfPopulation>();
-  const { year } = useParams();
 
   //WIP: using form context to get the labels for this component temporarily.
   const labels: any = useContext(SharedContext);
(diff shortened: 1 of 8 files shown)


services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -187,13 +189,15 @@ export const DefinitionOfPopulation = ({
...
         )}
       {childMeasure && ChildDefinitions(register)}
-      <CUI.Box my="5">
-        <QMR.TextArea
-          formLabelProps={{ fontWeight: "400" }}
-          label="If this measure has been reported by the state previously and there has been a change in the included population, please provide any available context below:"
-          {...register(DC.CHANGE_IN_POP_EXPLANATION)}
-        />
-      </CUI.Box>
+      {year && parseInt(year) < 2024 && (
```
ailZhou, line 192: Request: This does fit the ask but we want to cut down on using year conditionals as we never know when they want to turn this back on. There's a new 'labels' folder that contains the label text for each component by year, there should be an area for DefinitionsOfPopulation. If you can just add a new key (i.e. changeInPopExplain or anything really) for 2021 - 2023 and then not have it for 2024. You can use the existence of that key to determine whether to render that text or not. 

That object is already loaded through `const labels: any = useContext(SharedContext);` on line 173 so you can call it through here.

gmrabian (dismissed):
i'm good with the changes, pending Aileen's request

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
karla-vm: I've updated that, thanks for calling that out!

gmrabian (approved):

PR #2164, opened 2024-04-04
validation text change

britt-mo:
### Description
`Update the language in the validation that flags when a state selects "Other Data Source" but doesn't enter text in the box provided. Currently, the flag says "Data Source Error: You must select a data source" which may be confusing to states because a data source has already been selected. Can you please update the flag language to say "Data Source Error: Please describe the Other Data Source"? `

### Related ticket(s)

CMDCT-3480

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/measures/2023/shared/globalValidations/validateAtLeastOneDataSourceType/index.test.ts
+++ b/services/ui-src/src/measures/2023/shared/globalValidations/validateAtLeastOneDataSourceType/index.test.ts
@@ -45,6 +45,18 @@ describe("validateOneDataSourceType", () => {
     };
     errorArray = [...validateAtLeastOneDataSourceType(formData)];
     expect(errorArray.length).toBe(1);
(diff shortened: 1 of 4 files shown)


keeysnc (dismissed):

ailZhou (changes requested):
I think that message is shared between any data source that is selected, even non other data source. When I select: 

Screenshot 2024-04-09 at 9 17 37 AM

I get 2 error message back, one validation is run against administrative data (which makes sense for the original message) and the second for other data source (which is what the new message is for).
Screenshot 2024-04-09 at 9 17 44 AM

I think you might need a conditional to track what the label is and then return the correct text.

keeysnc (approved):

ailZhou (approved):
lgtm! tested in deploy link and the correct validation message is displaying for both other and non other.

PR #2173, opened 2024-04-10
CPU-AD Content Update and Additional Fields

ailZhou:
**Deploy Link**
https://d2h0t1dukdfa66.cloudfront.net/

### Description

For CPU-AD, they wanted to combine NDR sets and field sets under the same performance measure. That is not something normally done in QMR, its either all NDR or all single fields (i.e. PCR-AD, PCR-HH). The way QMR handles it is to create a new PerformanceMeasure and use that over the default one. 

Things changed:
- New PerformanceMeasure.tsx added to CPU-AD. This is the override file for this measure. It handles generating the correct fields under performance measure and routes to a rate calculations if necessary.
- Added new key in LabelData `isField`. This helps determine which qualifier is using an NDR set, or is a field. This may get updated during a future refactor.
- The UI does not look exactly like the ticket but that's okay, David had approved of the fields being underneath the label and not to the side.
- This is the data structure of the captured data:
Screenshot 2024-04-10 at 2 01 54 PM

### Related ticket(s)

CMDCT-3388

---
### How to test

1) Sign into QMR, any user
2) Select adult core set, go into CPU-AD
3) In Measure Specification, select the first radio button to enable the performance measure section
4) Scroll down and you should see something like this (but without it filled in):
Screenshot 2024-04-10 at 12 57 18 PM
5) Fill out the fields and validate
6) Run local dynamo db: `DYNAMO_ENDPOINT=http://localhost:8000 dynamodb-admin`
7) Check to make sure that the saved data looks similar to this:
Screenshot 2024-04-10 at 2 01 54 PM

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


gmrabian (approved):
deployed env and db look good

PR #2176, opened 2024-04-11
Refactor AddCoreSetCards.tsx 

ailZhou:
**Deploy Link**
https://d4kbxh9soszcy.cloudfront.net/

### Description

The generation of the add coreSet buttons are now being controlled in coreSetByYear.tsx file (including the text blob).
Screenshot 2024-04-15 at 9 36 26 AM

Here's the data chunk to correlate:
```
"2023": [
    {
      type: "coreSet",
      label: "Adult",
      loaded: true,
      abbr: ["ACS"],
    },
    {
      type: "coreSet",
      label: "Child",
      path: "add-child",
      abbr: ["CCS", "CCSC", "CCSM"],
    },
    {
      type: "coreSet",
      label: "Health Home",
      path: "add-hh",
      abbr: ["HHCS"],
      stateList: getHHStates("2023"),
    },
    {
      type: "text",
      label:
        "Only one group of Adult Core Set Measures can be submitted per FFY",
    },
  ],
```
**So what changed?**
1) **spaLib.ts** - This file wasn't suppose to be part of this PR but some spa ids did not get removed correctly last year, so its being removed now. This was the original source for determining whether a state will have a healthHome or not.
2) **coreSetByYear.tsx** - This file will now be what is used to control what coreSet should be in what state and for what year. Most keys should be self explanatory except `loaded`, this one is use to determine whether the coreSet is already loaded in the state by default, and therefore does not require an add core set button. This instead, means it is required to be in table already.
3) **AddCoreSetCards.tsx** - This component does not do any data formatting, only for rending a coreSet button or text label. The `GetSetCard` function is where any new render type will get added. It should reduce the noise in this script.
4) **StateHome/index.tsx** - This is where the data object in `coreSetByYear.tsx` is called and formatted, before it is rendered in `AddCoreSetCards.tsx`.

### Related ticket(s)

CMDCT-3448

---
### How to test

1) Sign into [stateuserCA@test.com](mailto:stateuserCA@test.com). 
2) Verify that the coreSet selection for each year looks like the screenshot below.

2021 & 2022  
Screenshot 2024-04-15 at 9 24 18 AM

2023 (Health Home was removed for CA during this year)
Screenshot 2024-04-15 at 9 24 34 AM

2024 (Adult coreset will now be added by button, that is why you can see it now but it will be disabled until everything is merged together in a different pr)
Screenshot 2024-04-15 at 9 24 45 AM

### Notes

3) In reporting year 2022, add a child core set and a health home coreset (there is two in the dropdown, only add one). [Add Child Core Set] button should become inactive but Health Home should still be active. 
4) Add the second health home and the button should become inactive.
5) Delete the child core set, click the 3 dot button on the right and select [Delete]
6) The button should be enable again.
7) This is sufficient to test if the data is being pulled correctly from coreSetByYear.tsx file. You can also go into that file and add a coreset or remove one just to see how it works but that is not necessary for this test.

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/libs/spaLib.ts
+++ b/services/ui-src/src/libs/spaLib.ts
@@ -254,20 +254,9 @@ export const SPA: { [year: string]: SPAi[] } = {
   ],
   "2023": [
     {
-      id: "19-0037",
-      state: "CA",
-      name: "California Health Home Program (HHP)",
-    },
-    { id: "20-0002", state: "CA", name: "California Behavioral Health Home" },
-    {
       id: "15-014",
       state: "CT",
       name: "Serious and Persistent Mental Illness",
-    },
-    {
-      id: "18-006",
-      state: "DE",
-      name: "Delaware Assertive Community Integration Support Team (ACIST) Health Home",
     },
     {
       id: "18-0006",
@@ -334,8 +323,6 @@ export const SPA: { [year: string]: SPAi[] } = {
       name: "Health Home for High-Cost, High-Needs Enrollees ",
     },
     { id: "22-0073", state: "NY", name: "New York I/DD Health Home Services" },
-    { id: "14-0012", state: "OK", name: "Oklahoma Health Home (Adults)" },
-    { id: "14-0011", state: "OK", name: "Oklahoma Health Home (Children)" },
     {
       id: "18-009",
       state: "RI",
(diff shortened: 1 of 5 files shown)


britt-mo (approved):
Tested per instructions, everything appeared and functioned as expected

PR #2184, opened 2024-04-16
Remove cypress steps for tests that have been deleted

gmrabian:
### Description

Completing work in removing cypress tests by deleting the steps to run them in CI

### Related ticket(s)

CMDCT-3379
CMDCT-3380
CMDCT-3381
CMDCT-3382

---
### How to test

Cypress steps for adult, child, health home, and a11y don't try to run

### Notes

We'll need to make sure to add something like this back when we add new tests as part of CMDCT-3529

---
### Pre-review checklist

- [ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [ ] I have updated relevant documentation, if necessary
- [ ] I have performed a self-review of my code
- [ ] I have manually tested this PR in the deployed cloud environment

PR #2188, opened 2024-04-17
Assert failure on matrix when test files not present

berryd:
### Description
This gave me a mild panic attack yesterday....

I found some Snyk PRs that were automerged that were indicated to have failed tests.  After playing with this quite a bit, it seems that the failures for the matrix are produced asynchronously, so a build will go green when the matrix runners fail to build, then go red a few seconds later, which was enabling PRs to merge in that interim, and in the case of Snyk PRs, they are set to auto-merge when all checks pass, so this little blip let them slip through.  

This little change just asserts that the files are identified before the cypress-workflow passes off the config to the deploy workflow.  An example of it successfully failing can be seen here:  [deploy workflow](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/8710353652)

### Related ticket(s)

There isn't a ticket, but it goes hand in hand with this:  [Remove cypress steps for tests that have been deleted ](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2184)

---
### How to test

See the above mentioned failure to see this succeeding.  I have since merged the changes from the PR referenced above so this build should succeed as expected, assuming it cleans up the removed PR checks...

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/.github/workflows/cypress-workflow.yml
+++ b/.github/workflows/cypress-workflow.yml
@@ -35,10 +35,9 @@ jobs:
       - uses: actions/checkout@v4
       - id: set-test-matrix
         run: |
-          TEST_PATH="tests/cypress/cypress/e2e/${{ inputs.test-path }}/*"
-          [ -n "$(find ${TEST_PATH} -maxdepth 0 -type d -empty 2>/dev/null)" ] && echo "No test definition found...exiting" && exit 1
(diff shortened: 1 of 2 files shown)


berryd:
This will never pass checks because the removed workflows are still blocking merging.  I'm not sure why that is the case since I merged master into this and destroyed the original branch, created a new one, and a new PR.  Anyway, I'm standing up another PR to see if this is still ongoing with an empty commit elsewhere.

gmrabian (approved):

PR #2201, opened 2024-04-24
[WIP]: enable idempotency in all serverless services

gmrabian:
### Description

https://github.com/stratiformdigital/serverless-idempotency-helper

Plugin only helps with services that deploy functions, because they compare the existing packaged functions to the latest changes. In QMR the following services have functions:
- app-api
- uploads
- ui-auth

This is half of the total services. 

Speed tests
[First and second run of a standard branch](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/8788877588) (first being no existing resources, second being no-change re-deploy)
We can see the [ui-auth service takes 42s](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/8788877588/job/24266397576#step:11:155) to determine there are no changes and skip
The [ui-auth service with the idempotency plugin takes the same amount of time](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/8837106060/job/24266321059#step:11:115)

| | w/o helper | with helper (on some) |
|----|----|-----|
| Clean deploy | 23:23 | 31:39  |
| Re-deploy no changes | 8:51  |  6:39 |
| Re-deploy no changes (2) | 6:48 | 9:12  |

### Related ticket(s)

CMDCT-3447

---
### Notes

---
### Pre-review checklist


gmrabian:
> Do you happen to know if this plugin was developed before serverless had idempotency itself? I’m wondering if the framework has since added similar functionality and this plugin is redundant or deprecated as a result (again, just trying to guess at the variance with and without the plugin)

Berry had this to say
> It's OBE.  It was to correct an issue with webpack producing seemingly different artifacts from identical sources.

As a result I think we can close this PR, being that we didn't see significant benefit to using the plugin, and it seems to no longer be relevant

PR #2206, opened 2024-04-29
Improve artifact upload

berryd:
### Description
QMR differs in many ways from the other apps, so some of the issues with artifact naming convention collisions won't exit, but there is still some minor opportunity for improvement, and in the course of this work, I will verify that the uploads do in fact work.  

One thing specifically different in QMR than other applications is that e2e tests run concurrently in a GHA matrix.  This produces an interesting problem of a race condition being able to override a previously saved artifact should more than one runner fail.  This problem is mitigated by prefixing the artifact with the name of the matrix (the matrix is a singleton because it determines the path to the test to execute)

### Related ticket(s)

CMDCT-3556

---
### How to test

See [this run](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/8882953583) for an example.  For this instance, I forced two of the e2e tests to fail to demonstrate that the artifacts could be successfully saved.  A successful run will produce no artifacts.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/.github/workflows/cypress-workflow.yml
+++ b/.github/workflows/cypress-workflow.yml
@@ -73,7 +73,7 @@ jobs:
         run: |
           pwd
           ls -alh 
-          find . -iname screenshots -type d  
+          find tests -iname \* -type d  
       - name: Upload screenshots
         uses: actions/upload-artifact@v4
         #if: failure()
--- a/.github/workflows/cypress-workflow.yml
+++ b/.github/workflows/cypress-workflow.yml
@@ -77,6 +77,6 @@ jobs:
         #if: failure()
         if: always()
         with:
-          name: cypress-test-results
+          name: ${{ matrix.containers }}-test-results
           path: tests/cypress/screenshots/
           retention-days: 14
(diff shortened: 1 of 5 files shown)


.github/workflows/cypress-workflow.yml
```diff
@@ -69,9 +69,13 @@ jobs:
           CYPRESS_STATE_USER_4: ${{ secrets.cypress-user4 }}
           CYPRESS_ADMIN_USER: ${{ secrets.cypress-admin-user }}
           CYPRESS_QMR_PASSWORD: ${{ secrets.cypress-password }}
+      - name: How can I make this unique
+        run: |
+          cat $GITHUB_ENV
```
gmrabian, line 74: meant to stay?

gmrabian (approved):
good except for one lingering "debug" step (?)

.github/workflows/cypress-workflow.yml
berryd: nope

berryd: Cleanup pushed

gmrabian (approved):

PR #2212, opened 2024-05-03
Add "View Combined Rates" button to state home page

gmrabian:
### Description

- Add route for combined-rates
- Add button to state home page to view combined rates
- Update styling (open to any better ideas here!)
- Add tests

### Related ticket(s)

CMDCT-3454

---
### How to test

Deployed url: https://d313thmob07q2p.cloudfront.net/
- Log in as any state user
- Verify the button "View Combined Rates" shows up below the year dropdown
- Change to any year other than 2024 and verify the button goes away
- Shrink the view to mobile size and make sure the button looks okay
- Check views against [Figma](https://www.figma.com/file/AJweEUn95IZFAN3ldIcZDH/MDCT---General?type=design&node-id=2666-2508&mode=design&t=NrRav96Mrll5wRT4-4)

---
### Pre-review checklist

---
#### Review

### DIFF
--- a/services/ui-src/src/views/StateHome/index.tsx
+++ b/services/ui-src/src/views/StateHome/index.tsx
@@ -19,6 +19,7 @@ import { BannerCard } from "components/Banner/BannerCard";
 import { coreSets, CoreSetField } from "shared/coreSetByYear";
 
 import { useFlags } from "launchdarkly-react-client-sdk";
+import { Link } from "react-router-dom";
 
 interface HandleDeleteData {
   state: string;
@@ -59,8 +60,8 @@ const ReportingYear = () => {
   }
 
   return (
-    <CUI.Box w={{ base: "full", md: "48" }}>
-      <CUI.Text fontSize="sm" fontWeight="600" mb="2">
+    <CUI.Box w={{ base: "full", md: "48" }} display={{md: "flex"}} flexDirection={{md: "column"}} alignItems={{md: "flex-end"}}>
+      <CUI.Text fontSize="sm" fontWeight="600" alignSelf={{md: "flex-start"}}>
         Reporting Year
       </CUI.Text>
       <CUI.Select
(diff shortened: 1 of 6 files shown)


ailZhou (approved):
lgtm! testing on deploy. button shows up only for 2024 and goes to a new empty page. 

QMR does have a theme.ts file but that doesn't look to be heavily used, so I think the way you added css is within the current pattern.

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -80,6 +86,29 @@ const ReportingYear = () => {
           </option>
         ))}
       </CUI.Select>
+      {year === "2024" && (
```
rocio-desantiago, line 89: curious why we hard code the year. Will we redo this for 2025 next year?

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -88,11 +117,11 @@ const Heading = () => {
   const { year } = useParams();
   return (
     <CUI.Box display={{ base: "block", md: "flex" }}>
-      <CUI.Box maxW="3xl" pb="6">
+      <CUI.Box maxW="3xl" pb="6" pr={{ md: "4rem" }}>
```
rocio-desantiago, line 120: I'll have to familiarize myself with CUI.Box. Not familiar with pb, pr, py, pt yet

rocio-desantiago (approved):

services/ui-src/src/views/StateHome/index.tsx
gmrabian: We try to avoid this where we can, but in this case this button only shows up for that reporting year. Since it's a shared component we need to check. I suppose we could do `>= "2024"` but we don't yet know about 2025 yet 🤔

PR #2224, opened 2024-05-08
Upgrade QMR to Node 20

gmrabian:
### Description

- Upgrade `.nvmrc` to node 20
- Upgrade services to node 20
- Upgrade dependencies that need it for compatibility
- Remove redundant dependencies

Note: there is a lambda that isn't on Node 20. It's created by serverless to link resources as a workaround for...itself. We might be able to narrow down where it comes from with an infrastructure audit, but it just runs once on deploy and then sits there. It'll only be a concern if AWS removes support for Node16, and in that case everyone using serverless will need a solution, so they should upgrade it by then (right...?)

### Related ticket(s)

CMDCT-2659

---
### How to test

Deployed url: https://d1zh5vte8mrjmn.cloudfront.net/
Everything works as expected
Lambdas don't fail

---
### Pre-review checklist


.nvmrc
```diff
@@ -1 +1 @@
-v18.20.2
+v20.11.1
```
rocio-desantiago, line 1: does this cover `^20.12.7`?

package.json
```diff
@@ -25,7 +25,7 @@
     "@serverless/compose": "^1.3.0",
     "@types/dotenv": "^8.2.0",
     "@types/flat": "^5.0.2",
-    "@types/node": "^18.19.31",
+    "@types/node": "^20.12.7",
```
rocio-desantiago, line 28: does this version work with .nvmrc having `v20.11.1`?

rocio-desantiago (approved):
Question about `.nvmrc` file vs `package.json`

package.json
gmrabian: see above

.nvmrc
rocio-desantiago: Sweet :shipit:

ailZhou (approved):

PR #2226, opened 2024-05-08
Improve destroy

berryd:
### Description
All of our applications have been leaking infrastructure due to the reliability of the destroy mechanism.  In the past it's been a rather manual process to clean up latent and orphaned infrastructure in order to keep account costs under control.  Luckily, our ephemeral environments and very light weight and cheap to operate, but manually cleaning up orphaned infrastructure doesn't scale.

Luckily, this problem has been solved.  The purpose of this PR is retrofit an existing solution into QMR as a trial to identify what is needed to implement this across the board.

Notable changes:
- The dev script (executed by ./dev), has been renamed to run (executed by ./run)
  - The previous commands are unaffected so `./run local` is equivalent to prior functionality
- The destroy.sh script has been removed
- Issuing destroy locally is accommodated by executing `./run destroy --stage [stage]`

### Related ticket(s)

CMDCT-1929

---
### How to test
See this feature working correctly in this [job](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/9038889075/job/24840810586).  Note that the debug step has since been removed.

### Notes
Before this branch is merged in, this workflow needs to be manually executed.  This can be accomplished by the following:
- Navigate to the destroy action
- Select `run workflow`
- Select `improve-destroy` branch
- Enter `improve-destroy` in the text field
- Trigger the workflow and wait for execution
- Merge branch

While trying to debug some issues, I discovered a hack to enable outputting secrets to a GHA workflow.  This can be accomplished similarly to the following 
```
      - name: DEBUG
        run:  echo ${{ env.AWS_DEFAULT_REGION }}  | sed 's/./& /g'
```

### Additional Notes
The new destroy script allows for additional filters which are based on Resource Tags.  It would be a good idea for us to implement additional resource tags and I intend to make a follow on ticket to do so.  These tags allow for additional namespacing such as aggregating resources by service, not just Cloudformation stack, and namespacing Projects which would allow for multiple products to be deployed to a single account without collisions.  

The `deploy.sh` script can also be deprecated putting all local and CI commands within the single `run` wrapper.  Concerns were expressed with the MACPRO implementation of the runner, so I refactored it to not use wrapped anonymous functions in the yargs workflow.  

---
### Pre-review checklist

---
#### Review

#### Security

---


gmrabian (approved):
Changes here look good. I didn't do a deep inspection of the stratiform serverless destroyer but I'm ok with this going in

PR #2229, opened 2024-05-10
Cypress Dropdown Fix

ailZhou:
### Description

Due to the addition of a deletable Adult Core Set, the cypress test have started to fail. The delete code we were using only had to worry about one core set per user as there would only be one set of deletes in the system. Because adult measures have a delete option now, if both adult and child cypress is run at the same time and is generated at the same time, it creates an issue.

We were using:  `cy.get('[data-cy="Delete"]').first().click({ force: true });` to delete the first in the stack order, but this is indifferent to whether the button is visible or not. So it didn't matter if the child menu or the adult menu was made visible, we would still delete the adult measure as that menu still exist even if it is hidden; it will usually be the first in the stack. 

Screenshot 2024-05-10 at 12 53 35 PM

Another issue is that things would be animating while trying to click reset / complete all the measure buttons, which would cause misfires to the wrong button in the menu. So this code was added to some of the buttons
```
{
      force: true,
      waitForAnimations: false,
}
```

### Related ticket(s)

N/A

---
### How to test

If the cypress test passes, we should be good.

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/tests/cypress/support/commands.ts
+++ b/tests/cypress/support/commands.ts
@@ -145,25 +145,31 @@ Cypress.Commands.add("displaysSectionsWhenUserNotReporting", () => {
   ).should("be.visible");
 });
 
-// helper recursive function to remove added core sets
-const removeCoreSetElements = (kebab: string) => {
+const clickCoreSetAction = (kebab: string, action: string) => {
   cy.get('[data-cy="tableBody"]').then(($tbody) => {
     if ($tbody.find(kebab).length > 0) {
       cy.get(kebab).first().click();
       cy.wait(3000);
-      cy.get('[data-cy="Delete"]').first().click({ force: true });
-      cy.get('[data-cy="delete-table-item-input"]').type("delete{enter}");
-      cy.wait(3000);
-      removeCoreSetElements(kebab);
+      cy.get(action).click();
     }
   });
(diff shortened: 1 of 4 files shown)


tests/cypress/support/commands.ts
```diff
@@ -145,34 +145,48 @@ Cypress.Commands.add("displaysSectionsWhenUserNotReporting", () => {
   ).should("be.visible");
 });
 
-// helper recursive function to remove added core sets
-const removeCoreSetElements = (kebab: string) => {
+const clickCoreSetAction = (kebab: string, action: string) => {
```
gmrabian, line 148: `action` is a little misleading. this is just a selector, right?

gmrabian (approved):
this resolves the issues I saw with the tests

tests/cypress/support/commands.ts
ailZhou: true, but we call that column core set actions 🤔

ailZhou: i am willing to rename this, but in a different PR because i don't want to wait for this to compile again

BearHanded (approved):

PR #2239, opened 2024-05-17
[QMR] All Child and Adult Measures - In line warning banner (Data Source)

ajaitasaini:
### Related ticket(s)

CMDCT-3628

---
### How to test

- Check any adult or child measure in FFY2024, click "Other Data Source" and confirm that a banner is showing up correctly 
- Make sure the banner is not showing up for any health home measure or any measure before 2024
- special measures to double check: CPA-AD, MSC-AD, CPC-CH

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


ailZhou:
The locations of where you placed the alerts looks to be correct to me and they are showing up. There's just two areas that could use a second look.
1) Based on the ticket, I think only Other Data Source is suppose to have the warning and not Administrative Data Other 
Screenshot 2024-05-20 at 12 45 34 PM
2) This is showing up for all the years but I think we only want this for reporting year 2024 and onward. An example of how we handle only rendering something for a year and not the other can be seen in `OptionalMeasureStrat/index.ts`. If the text key exist in the commonQuestionsLabel.tsx for that year, it will render, if not, it will ignore it.

Screenshot 2024-05-20 at 12 53 13 PM

Let me know if anything was confusing or you want a better explanation of anything!

benmartin-coforma:
I think we'll also need to hit CPA-AD specifically; it has its own version of this question, just like MSC-AD

services/ui-src/src/measures/2024/MSCAD/questions/DataSource.tsx
```diff
@@ -31,6 +33,16 @@ export const DataSource = () => {
                 }
                 {...register("DataSource-CAHPS-Version-Other")}
               />,
+              <CUI.Box mt="8">
+                <Alert heading="Please Note" variation="warn">
+                  <p className="ds-c-alert__text">
```
keeysnc, line 38: I think you can use CUI.Text for this instead of p tag for consistency purposes.

services/ui-src/src/shared/commonQuestions/DataSource.tsx
```diff
@@ -79,7 +80,16 @@ const buildDataSourceOptions: DSCBFunc = ({ data = [], parentName }) => {
           label={parseLabelToHTML(node.hint!)}
           name={`${DC.DATA_SOURCE_SELECTIONS}.${adjustedParentName}.${DC.DESCRIPTION}`}
           key={`${DC.DATA_SOURCE_SELECTIONS}.${adjustedParentName}.${DC.DESCRIPTION}`}
-        />
+        />,
+        <CUI.Box mt="8">
+          <Alert heading="Please Note" variation="warn">
+            <p className="ds-c-alert__text">
```
keeysnc, line 86: Same suggestion for the p tag

services/ui-src/src/shared/commonQuestions/DataSourceCahps.tsx
```diff
@@ -44,6 +46,16 @@ export const DataSourceRadio = ({ data = defaultData }: DataSourceProps) => {
                   fontSize: "normal",
                 }}
               />,
+              <CUI.Box mt="8">
+                <Alert heading="Please Note" variation="warn">
+                  <p className="ds-c-alert__text">
```
keeysnc, line 51: Same suggestion for the p tag

services/ui-src/src/measures/2024/CPAAD/questions/DataSource.tsx
```diff
@@ -32,6 +40,16 @@ export const DataSource = () => {
                 }
                 {...register("DataSource-CAHPS-Version-Other")}
               />,
+              (coreset === "adult" || coreset === "child") &&
```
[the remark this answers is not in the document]
ajaitasaini: that's a good point, and i guess this check is not doing anything. will remove!

services/ui-src/src/shared/commonQuestions/DataSourceCahps.tsx
```diff
@@ -6,15 +6,23 @@ import { DataSourceData, defaultData } from "../types/TypeDataSourceCahps";
...
   data?: DataSourceData;
+  adultMeasure?: boolean;
+  childMeasure?: boolean;
 }
 
 /**
  * Fully built DataSource component
  */
-export const DataSourceRadio = ({ data = defaultData }: DataSourceProps) => {
+export const DataSourceRadio = ({
```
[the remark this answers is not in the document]
ajaitasaini: I believe I added this in case we would end up using DataSourceRadio for a 2024 or beyond measure, but I was not too sure about this myself. I can follow up with Aileen and if it looks unnecessary then I can scrap the changes for this file!

services/ui-src/src/measures/2024/AABAD/index.tsx
```diff
@@ -44,7 +44,7 @@ export const AABAD = ({
         <>
           <CMQ.StatusOfData />
           <CMQ.MeasurementSpecification type="HEDIS" />
-          <CMQ.DataSource />
+          <CMQ.DataSource coreset="adult" />
```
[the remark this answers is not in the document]
ajaitasaini: 1. yes we can match the `DateRange` pattern!
2. there is no existing prop from what I know, and other than the pattern `DateRange` follows, the only other similar example we noticed was passing in `healthHomeMeasure={true}` or `childMeasure={true}` for DefinitionOfPopulation. We originally were passing in a prop in a very similar pattern, and then moved to `coreset = "_____"`, etc so it was more reusable as a prop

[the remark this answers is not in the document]
gmrabian: I agree pathname isn't great. I think 1. is the best option, matching it with the DateRange pattern. Or maintaining `coreset` if that's a pattern you are starting to use widely.

Thanks for hearing me out!

karla-vm (dismissed):

services/ui-src/src/measures/2024/WCCCH/index.tsx
```diff
@@ -41,7 +41,7 @@ export const WCCCH = ({
         <>
           <CMQ.StatusOfData />
           <CMQ.MeasurementSpecification type="HEDIS" />
-          <CMQ.DataSource data={PMD.dataSourceData} />
+          <CMQ.DataSource data={PMD.dataSourceData} type="adult" />
```
gmrabian, line 44: this is a child measure, no?

gmrabian: looks like the DateRange value below is wrong as well

gmrabian (changes requested):

gmrabian (approved):

karla-vm (approved):
tested in deployed instance, looks good!

PR #2264, opened 2024-06-05
Combine Rates - Administrative Data Source Only Calculation 

ailZhou:
### Description

Huge thanks to @benmartin-coforma & @BearHanded for helping me fix the issues in this PR.

This PR contains a new rates table that we are using to store the new rate calculation data. The flow is that when a user validates or submits a measure, it will trigger the rate calculations after saving the measure data.

**Default calculations**:  (numerator / denominator) * 100;

**Measures with unique calculations**:
AAB-AD: (1 - numerator / denominator) * 100;
AMB-AD: (numerator / denominator) * 1000;
PQI01, PQI05, PQI08, PQI15:  (numerator / denominator) * 100000;

### Related ticket(s)

CMDCT-3651

---
### How to test

Run local db: `DYNAMO_ENDPOINT=http://localhost:8000 dynamodb-admin`
1) Sign into QMR, any separate state user (ie. [stateuser1@test.com](mailto:stateuser1@test.com))
2) In Adult Chip, pick AMM-AD:
 - Fill out the measure, especially data source and performance measure sections 
 - Validate measure
4) Do the same for Adult Medicaid AMM-AD
5) Go to local db and check the local-rates table and you should see a single database entry for that measure
6) Click it and make sure that the data looks similar to this:
```
{
  "compoundKey": "WA2024ACSAMM-AD",
  "measure": "AMM-AD",
  "state": "WA",
  "data": [
    {
      "column": "CHIP",
      "dataSourceSelections": {
        "AdministrativeData0": {
          "selected": [
            "MedicaidManagementInformationSystemMMIS"
          ]
        }
      },
      "dataSource": [
        "AdministrativeData"
      ],
      "rates": {
        "JjQeAC": [
          {
            "category": "Effective Continuation Phase Treatment",
            "uid": "JjQeAC.g91VU9",
            "label": "Ages 18 to 64"
          },
          {
            "category": "Effective Continuation Phase Treatment",
            "uid": "JjQeAC.Gjknmo",
            "label": "Age 65 and older"
          }
        ],
        "DFukSh": [
          {
            "uid": "DFukSh.g91VU9",
            "label": "Ages 18 to 64",
            "category": "Effective Acute Phase Treatment",
            "rate": "50.0",
            "numerator": "1",
            "denominator": "2"
          },
          {
            "uid": "DFukSh.Gjknmo",
            "label": "Age 65 and older",
            "category": "Effective Acute Phase Treatment",
            "rate": "75.0",
            "numerator": "3",
            "denominator": "4"
          }
        ]
      }
    },
    {
      "column": "Medicaid",
      "dataSourceSelections": {
        "ElectronicHealthRecords": {}
      },
      "dataSource": [
        "ElectronicHealthRecords"
      ],
      "rates": {
        "JjQeAC": [
          {
            "category": "Effective Continuation Phase Treatment",
            "uid": "JjQeAC.g91VU9",
            "label": "Ages 18 to 64"
          },
          {
            "category": "Effective Continuation Phase Treatment",
            "uid": "JjQeAC.Gjknmo",
            "label": "Age 65 and older"
          }
        ],
        "DFukSh": [
          {
            "uid": "DFukSh.g91VU9",
            "label": "Ages 18 to 64",
            "category": "Effective Acute Phase Treatment",
            "rate": "50.0",
            "numerator": "1",
            "denominator": "2"
          },
          {
            "uid": "DFukSh.Gjknmo",
            "label": "Age 65 and older",
            "category": "Effective Acute Phase Treatment",
            "rate": "80.0",
            "numerator": "4",
            "denominator": "5"
          }
        ]
      }
    },
    {
      "column": "Combined Rate",
      "rates": [
        {
          "uid": "JjQeAC.g91VU9",
          "label": "Ages 18 to 64",
          "category": "Effective Continuation Phase Treatment",
          "rate": "",
          "numerator": "0",
          "denominator": "0"
        },
        {
          "uid": "JjQeAC.Gjknmo",
          "label": "Age 65 and older",
          "category": "Effective Continuation Phase Treatment",
          "rate": "",
          "numerator": "0",
          "denominator": "0"
        },
        {
          "uid": "DFukSh.g91VU9",
          "label": "Ages 18 to 64",
          "category": "Effective Acute Phase Treatment",
          "rate": "50.0",
          "numerator": "2",
          "denominator": "4"
        },
        {
          "uid": "DFukSh.Gjknmo",
          "label": "Age 65 and older",
          "category": "Effective Acute Phase Treatment",
          "rate": "77.8",
          "numerator": "7",
          "denominator": "9"
        }
      ]
    }
  ],
  "year": "2024",
  "lastAltered": 1718281781252
}
```

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


services/app-api/handlers/rate/calculateAndPutRate.ts
```diff
@@ -0,0 +1,66 @@
...
+) => {
+  const { coreSet, measure, state, year } = event!.pathParameters!;
+  const combinedTypes = [Types.CoreSetAbbr.ACS, Types.CoreSetAbbr.CCS];
+  const combinedCoreSet: Types.CoreSetAbbr = combinedTypes.find((type) =>
+    coreSet!.includes(type)
+  )!;
+
+  //only do the rate calculation if the measure is adult or child and is a split
+  if (
+    (coreSet!.length === 4 && combinedCoreSet === "ACS") ||
```
BearHanded, line 33: can enums for ACS and CCS can be used here

services/app-api/storage/table.ts
```diff
@@ -0,0 +1,59 @@
...
+import { getMeasure } from "../handlers/measures/get";
+
+const coreSetGroup = {
+  ACS: [Types.CoreSetAbbr.ACSC, Types.CoreSetAbbr.ACSM],
+  CCS: [Types.CoreSetAbbr.CCSC, Types.CoreSetAbbr.CCSM],
+  HHCS: [Types.CoreSetAbbr.HHCS],
+};
+
+export const putToTable = async (
+  event: APIGatewayProxyEvent,
```
BearHanded, line 16: This is not a handler and has no need for 90% of the content in the event. It can be more flexible by taking explicitly the arguments needed, and avoid having to fake an event when used in other contexts.

services/app-api/storage/table.ts
```diff
@@ -0,0 +1,59 @@
...
+      },
+      "post"
+    ),
+  };
+  await dynamoDb.update(params);
+  return { status: StatusCodes.SUCCESS, body: params };
+};
+
+export const getMeasureByCoreSet = async (
+  event: APIGatewayProxyEvent,
```
BearHanded, line 42: Same thought here about the event

services/app-api/handlers/rate/calculateAndPutRate.ts
```diff
@@ -0,0 +1,66 @@
...
+  ) {
+    const data = await getMeasureByCoreSet(event, combinedCoreSet!);
+    const formattedData: [] = formatMeasureData(data);
+
+    //based on the measure data, it will check against the calculations that exist and see which one is a match
+    const calculation: RateCalculation | undefined = dataSrcCalculations.find(
+      (cal) => cal.check(formattedData)
+    );
+
+    const combinedRates = [
```
BearHanded, line 44: So the plan here is to have an array with everything that feeds into the table, correct?

Alternatively the inputs to the calc can be stored in an array and the calculation result can be stored on a field in RateCalculation. Both work.

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
```diff
@@ -0,0 +1,75 @@
+import { RateCalculation } from "./rateCalculation";
+
+export class AdminstrativeCalculation extends RateCalculation {
+  dataSrcMap: any[] = [
+    {
+      Medicaid: ["AdministrativeData"],
```
BearHanded, line 6: Can these be enums?

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
```diff
@@ -0,0 +1,75 @@
...
+        const medicaidSrcExist = dataSrc.Medicaid.every(
+          (medSrc: string) => dataSources.Medicaid?.indexOf(medSrc) > -1
+        );
+        //if data source is a match in both CHIP & medicaid return true
+        if (chipSrcExist && medicaidSrcExist) {
+          return true;
+        }
+      }
+    }
+    return false;
```
BearHanded, line 56: What gets saved in the event that the calc is not valid? David mentioned the page still being visible for users when they are not using split measures

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  getFormula(measure: string): Function {
+    const abbr = measure.slice(0, 3);
+    switch (abbr) {
+      case "AMB":
```
BearHanded, line 61: do we need any comments about these special cases?

and can we use enums?

services/app-api/handlers/rate/calculateAndPutRate.ts
ailZhou: moved to enums!

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
ailZhou: if the combined rates is false, it stores an empty object {}

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
ailZhou: i'm not sure about special comments but i did move them to an enum

services/app-api/storage/table.ts
```diff
@@ -0,0 +1,59 @@
...
+) => {
+  const measures: any[] = [];
+  if (coreSet) {
+    const group = coreSetGroup[coreSet];
+
+    for (let i = 0; i < group.length; i++) {
+      const formatEvent = { ...event };
+      if (formatEvent?.pathParameters) {
+        formatEvent.pathParameters.coreSet = group[i];
+        const measure = await getMeasure(event, undefined);
```
BearHanded, line 53: calling a handler is probably not best practice

services/uploads/serverless.yml
benmartin-coforma, line 1: Are we planning on sending rateTable data upstream via S3 rather than Kafka, as we do with measure data? If yes then we will need more code & more external integration setup here... if no then we don't need to touch `services/uploads` at all.

ailZhou: sounds like the perfect question for @BearHanded!

services/app-api/storage/table.ts
ailZhou: removed calling the handler and used dynamo call instead

services/app-api/storage/table.ts
ailZhou: removed passing in APIGatewayProxyEvent and switching to using the path params

services/app-api/storage/table.ts
ailZhou: removed passing in APIGatewayProxyEvent and switching to using the path params

services/app-api/handlers/rate/calculateAndPutRate.ts
```diff
@@ -0,0 +1,66 @@
...
+  const combinedCoreSet: Types.CoreSetAbbr = combinedTypes.find((type) =>
+    coreSet!.includes(type)
+  )!;
+
+  //only do the rate calculation if the measure is adult or child and is a split
+  if (
+    (coreSet!.length === 4 && combinedCoreSet === "ACS") ||
+    combinedCoreSet === "CCS"
+  ) {
+    const data = await getMeasureByCoreSet(event, combinedCoreSet!);
```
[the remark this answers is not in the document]
ailZhou: removed passing in APIGatewayProxyEvent and switching to using the path params

services/uploads/serverless.yml
BearHanded: yea I think we'll need to support both. I'll create a ticket for it.

BearHanded: created

services/app-api/handlers/rate/get.ts
```diff
@@ -0,0 +1,22 @@
...
+    Key: {
+      compoundKey: dynamoKey,
+      measure: event!.pathParameters!.measure!,
+    },
+  };
+  const queryValue = await dynamoDb.get(params);
+  return {
+    status: StatusCodes.SUCCESS,
+    body: {
+      Item: queryValue,
```
benmartin-coforma, line 19: I hate to add this to the list of "things we'll have to change in order to break QMR of its `.Item` habit", but it does match our current conventions, so I think it's fine today.

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
benmartin-coforma, line 1: I think we will need to pay more attention to the edge cases in here, such as
* What if one of the measures is undefined?
* What if the data sources are undefined? What if the data sources are a string and not an array?
* What if the rates are undefined?
Right now the `?.` operators give us a sense of "well it probably won't crash", which is definitely good enough for today, but I'd like to see those edge cases addressed a little bit more definitely & explicitly.

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
```diff
@@ -0,0 +1,66 @@
...
+    coreSet!.includes(type)
+  )!;
+
+  //only do the rate calculation if the measure is adult or child and is a split
+  if (
+    coreSet.length === 4 &&
+    (combinedCoreSet === Types.CoreSetAbbr.ACS ||
+      combinedCoreSet === Types.CoreSetAbbr.CCS)
+  ) {
+    const data = await getMeasureByCoreSet(combinedCoreSet!, pathParameters);
```
benmartin-coforma, line 37: The logic here is to
1. Find the "parent" (combined) core set by substring
2. Confirm that we're in a separated core set by string length
3. Re-fetch our measure and fetch our "sibling" measure
   - which uses another map of combined core sets -> separated core sets, contained within table.ts
 
 I think we'll be able to simplify this to a single map of sibling -> sibling core sets, possibly using the data already in memory for our measure (since we just came from measures/update.ts).
 
But I am fine leaving that on the table as possible future re-work; this should get the job done as-is.

services/app-api/storage/table.ts
```diff
@@ -0,0 +1,62 @@
+import { convertToDynamoExpression } from "../handlers/dynamoUtils/convertToDynamoExpressionVars";
+import { StatusCodes } from "../utils/constants/constants";
+import dynamoDb from "../libs/dynamodb-lib";
+import * as Types from "../types";
+import { MeasureParameters } from "../types";
+
+export const putToTable = async (
```
benmartin-coforma, line 7: I'd prefer to see a less generic `putRate` function that knows its table name and the parameters required, taking that work away from its callers - essentially, a cousin of `getMeasureFromTable` below.

But I'd also like to see corresponding functions for getting/putting Measures, Core Sets, and Banners, which is definitely out of scope for this PR. I'm fine with putting off those kinds of changes to be done at the same (later) date.

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
```diff
@@ -0,0 +1,74 @@
...
+      arr.find((data) => data.column === "Medicaid")?.dataSource ?? [];
+
+    //if the user had selected hybrid as a data source, we will not use this calculation
+    const isHybrid = [chipSources, medicaidSources].some((srcs) => {
+      return (srcs as string[])?.includes(DataSource.Hybrid);
+    });
+
+    if (!isHybrid) {
+      for (var i = 0; i < this.dataSrcMap.length; i++) {
+        const dataSrc = this.dataSrcMap[i];
```
benmartin-coforma, line 41: `for..of` is generally preferable to both `for`, `for..in`, and `.forEach`, when possible. Which, here, it is 🤠 

```suggestion
      for (const dataSrc of this.dataSrcMap) {
```

services/app-api/storage/table.ts
```diff
@@ -0,0 +1,62 @@
...
+    for (let i = 0; i < group.length; i++) {
+      const measure = await getMeasureFromTable({
+        ...pathParams,
+        coreSet: group[i],
+      });
+      measures.push(measure);
+    }
+  }
+  return measures;
+};
```
benmartin-coforma, line 62: Normally I'd suggest `for..of` in place of this `for`, but here I would even go a step further, as long as we are getting 2 measures. If it were more I'd want to call `dynamoDb.query`, and if it were fewer we wouldn't need this function at all.

The type system should guarantee that coreSetGroup is always defined, although I should note that I have not run this code:

```suggestion
export const getMeasureByCoreSet = async (
  combinedCoreSet: "ACS" | "CCS",
  params: MeasureParameters
) => {
  const gets = coreSetGroup[combinedCoreSet].map(
    (coreSet) => getMeasureFromTable({ ...params, coreSet })
  );
  return Promise.all(gets);
};
```

services/app-api/types.ts
```diff
@@ -32,6 +43,22 @@ export interface Measure {
    * it is set on fetch, according to the metadata in measureList.ts.
    */
   autoCompleted?: boolean;
+  data?: {
+    DataSource?: DataSource[];
```
benmartin-coforma, line 47: This type is not fully correct; in addition to being an array of strings from the DataSource enum, DataSource can also be "CAHPS 5.1H". This will throw a bit of a wrench into the rest of our code, which so far doesn't account for that possibility.

```suggestion
    DataSource?: DataSource[] | "CAHPS 5.1H";
```

services/app-api/types.ts
```diff
@@ -32,6 +43,22 @@ export interface Measure {
...
+    DataSourceSelections?: unknown;
+    PerformanceMeasure?: {
+      rates?: {
+        [key: string]: StandardRateShape[];
```
benmartin-coforma, line 51: This type is not fully correct; in addition to being an array of StandardRateShapes, these values can be an array of what we might call FieldRateShapes. Once again this would throw a wrench into our code.

```ts
export type FieldsRateShape = {
  label: string;
  uid?: string;
  isTotal?: boolean;
  fields: {
    label: string;
    value?: string | number;
  }[];
};
```

```suggestion
        [key: string]: StandardRateShape[] | FieldsRateShape[];
```

services/app-api/types.ts
```diff
@@ -32,6 +43,22 @@ export interface Measure {
    * it is set on fetch, according to the metadata in measureList.ts.
    */
   autoCompleted?: boolean;
+  data?: {
+    DataSource?: DataSource[];
+    DataSourceSelections?: unknown;
```
benmartin-coforma, line 48: "unknown" here is accurate. I queried dev and prod to find all of the possible `DataSource` and `rate` shapes, but I didn't include DataSourceSelections in my investigation, so it is not known to me. This type definition will obviously have to change before we can use this field's values.

services/app-api/types.ts
```diff
@@ -32,6 +43,22 @@ export interface Measure {
    * it is set on fetch, according to the metadata in measureList.ts.
    */
   autoCompleted?: boolean;
+  data?: {
```
benmartin-coforma, line 46: There are a number of other properties that `data` can have, and it may be worthwhile to catalog them at some point. That would most certainly be out of scope today.

services/app-api/serverless.yml
```diff
@@ -204,6 +207,14 @@ functions:
           method: delete
           cors: true
           authorizer: ${self:custom.authValue.${self:custom.stage}, ""}
+  getRate:
+    handler: handlers/rate/get.getRate
+    events:
+      - http:
+          path: rate/{state}/{year}/{coreSet}/{measure}/get
+          method: get
```
benmartin-coforma, line 215: The `/get` here is in line with the QMR convention, but it is redundant with the `method`, and the convention has already been bucked by the banner endpoints. I'd recommend changing it, but I'd also recommend changing all the other endpoints... once again, a today vs. someday tradeoff to make.

```suggestion
          path: rate/{state}/{year}/{coreSet}/{measure}
          method: get
```

services/ui-src/src/libs/api.ts
```diff
@@ -141,6 +141,16 @@ async function deleteCoreSet(inputObj: any) {
   );
 }
 
+async function getRate(inputObj: any) {
+  const opts = await requestOptions();
+  opts.body = inputObj.body;
+  return API.get(
+    "coreSet",
+    `/rate/${inputObj.state}/${inputObj.year}/${inputObj.coreSet}/${inputObj.measure}/get`,
```
benmartin-coforma, line 149: If we were to change the path in app-api/serverless.yml, we would of course need to change it here too.

```suggestion
    `/rate/${inputObj.state}/${inputObj.year}/${inputObj.coreSet}/${inputObj.measure}`,
```

benmartin-coforma (approved):
I have a million comments below, which fall into two categories:
1. Nitpicks that are your call for whether to address them
2. Wistful imaginings of future days, safe for you to ignore.

As such, I approve. Very nice work!

BearHanded (approved):

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
```diff
@@ -0,0 +1,72 @@
...
+      CHIP: [DataSource.EHR],
+    },
+    {
+      Medicaid: [DataSource.EHR],
+      CHIP: [DataSource.Administrative],
+    },
+    {
+      Medicaid: [DataSource.Administrative, DataSource.EHR],
+      CHIP: [DataSource.Administrative],
+    },
```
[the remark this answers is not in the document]
ailZhou: made a ticket to figure out the strictness of the data source matching

services/app-api/serverless.yml
ailZhou: added it to a ticket as a global issue to look into fixing for all the paths

services/ui-src/src/libs/api.ts
ailZhou: added it to a ticket as a global issue to look into fixing for all the paths

services/app-api/handlers/rate/calculations/rateCalculation.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const uid = flattenRates
+      .filter(
+        (value, index, array) =>
+          array.findIndex((item) => item.uid === value.uid) === index
+      )
+      .map((item) => item.uid);
+    const sumRates = uid.map((id) =>
+      flattenRates.filter((rate) => rate.uid === id)
+    );
```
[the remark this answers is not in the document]
ailZhou: made a ticket for updating the tsconfig compile target: https://jiraent.cms.gov/browse/CMDCT-3730

services/app-api/handlers/rate/calculations/types.ts
```diff
@@ -0,0 +1,24 @@
+import { Measure, Program } from "../../../types";
+
+export enum DataSource {
+  Administrative = "AdministrativeData",
+  EHR = "ElectronicHealthRecords",
+  Hybrid = "HybridAdministrativeandMedicalRecordsData",
```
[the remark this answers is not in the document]
ailZhou: These will be added as we add the other calculations so saving it for next time

services/app-api/types.ts
ailZhou: This is something that we'll expand when we work on the other data source calculations as there will be other areas where we might need to pull the data.

services/app-api/handlers/measures/update.ts
```diff
@@ -55,5 +57,13 @@ export const editMeasure = handler(async (event, context) => {
   };
   await dynamoDb.update(params);
 
+  //in 2024 and onward, we added a new feature called combined rates which requires rate calculations to the rates table
+  if (parseInt(event!.pathParameters!.year!) >= 2024) {
+    //after updating the database with the latest values for the measure, we run the combine rates calculations for said measure
+    await calculateAndPutRate(
+      event!.pathParameters! as unknown as MeasureParameters
+    ); // TODO, fix this ugly double cast
```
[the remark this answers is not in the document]
ailZhou: Made a ticket to track fixing it: https://jiraent.cms.gov/browse/CMDCT-3732

services/app-api/handlers/rate/get.ts
ailZhou: Added it to a ticket because this is a whole can of worms that can break a lot of things: [CMDCT-3733](https://jiraent.cms.gov/browse/CMDCT-3733)

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
ailZhou: things are crashing, that's good enough for me! 🫥

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
ailZhou: highly agree that this can have way better logic. will note to fix up next week

benmartin-coforma (approved):
Nice!

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
ailZhou: saving this comment to think through next week

services/app-api/storage/table.ts
ailZhou: Fair point, we should expand because we hand an issue in get coreset where we are using the handler when we should probably just do a dynamo call. added to ticket: [CMDCT-3734](https://jiraent.cms.gov/browse/CMDCT-3734)

services/app-api/types.ts
ailZhou: going to fix that during the final compilation of all the code for combine rates

services/app-api/handlers/rate/calculations/types.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface FormattedMeasureData {
+  column: Program;
+  dataSource: NonNullable<NonNullable<Measure["data"]>["DataSource"]>;
+  dataSourceSelections: NonNullable<
+    NonNullable<Measure["data"]>["DataSourceSelections"]
+  >;
+  rates: NonNullable<
+    NonNullable<NonNullable<Measure["data"]>["PerformanceMeasure"]>["rates"]
+  >;
+}
```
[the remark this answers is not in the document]
ailZhou: Added it to a ticket as reminder: [CMDCT-3736](https://jiraent.cms.gov/browse/CMDCT-3736)

PR #2268, opened 2024-06-06
[QMR] Combined Rates Measurement Page - Initial

ajaitasaini:
### Description

This page will display the combined rates for a QMR user that are associated with the separated or combined measures. This story only covers the initial sections (Header, Copy, Links back to the measures).

### Related ticket(s)

CMDCT-3637

---
### How to test

1. In FFY 2024, go to the "View Combined Rates" section 
2. Click into one of the measures and make sure that the layout looks correct, specifically the header and a section with h2 and two links, one that redirects to the chip coreset measure page and another to the medicaid coreset measure page. Confirm that the links are redirecting to the correct pages.  
3. The body text is a placeholder and there's a followup ticket that will address that. 

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/Routes.tsx
+++ b/services/ui-src/src/Routes.tsx
@@ -128,6 +128,7 @@ export function useMeasureRoutes(): MeasureRoute[] {
               path: `:state/${year}/combined-rates/:measure`,
               element: (
                 <CombinedRatesMeasure
+                  measureName={foundMeasureDescription}
                   year={year}
                   measureId={measure}
                 ></CombinedRatesMeasure>
--- a/services/ui-src/src/Routes.tsx
+++ b/services/ui-src/src/Routes.tsx
@@ -124,8 +124,8 @@ export function useMeasureRoutes(): MeasureRoute[] {
             });
             //creating the routes for the combined rates measures
             measureRoutes.push({
-              key: `:state/${year}/combined-rates/:measure`,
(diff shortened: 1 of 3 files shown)


benmartin-coforma:
I like the code here, but navigating to the individual measures through the links on this page puts the app in a weird state, such that the `<` link on the breadcrumb trail doesn't work quite right.

I haven't dug into this, so I'm not sure how to explain it yet; screen recording below.

I would be surprised if this issue were related to the current work. More likely, it is an underlying issue with how QMR handles direct URL navigation. As such, it's probably out of scope here; I'm just making the note here so that we can follow up.

https://github.com/Enterprise-CMCS/macpro-mdct-qmr/assets/126210497/889fa290-2e99-46b8-9005-8dd747c41dc7

benmartin-coforma:
Cloudfront env is here: https://d3fybpu2xevre0.cloudfront.net/

ailZhou (approved):
LGTM, links are being generated and being redirected correctly to the measures. Also, thanks for fixing the issue with AAB-AD! 

Question: Not a blocker, but I'm wondering why the Link to Medicaid has a class tag but CHIP doesn't?

keeysnc (approved):

ajaitasaini:
@ailZhou good question r/e class tag! that's some leftover code that I will get rid of in a follow up PR for this component

PR #2279, opened 2024-06-10
CMDCT-3689: adding new definition of denominator options for various coresets

angelaco11:
### Description
According to [this](https://jiraent.cms.gov/browse/CMDCT-3689) ticket, for 2024 options, the definition of denominator needed to be changed to be different for each core set. I solved this by updating the regular `StandardDefinitions` components to be split into 2 components:
    - `StandardDefinitions` (regular ones)  and `CoreSetSpecificDefinitions` (new definitions for 2024)

The `StandardDefinitions` and `ChildDefinitions` components will either render the regular options, or it will render the core set specific definitions based on a boolean value I added to the 2024 labels called `useCoreSetSpecificOptions`. This allows us to render the new definitions that we need for 2024 or render the old definitions for any other year if that boolean does not exist in the other year labels.

---
### How to test
To test:
1. Navigate to each of the different core sets: Adult medicaid, Adult CHIP, Child Medicaid, Child CHIP, Adult not separate, Child not separate, Health home (you can find the last 3 with `stateUserDC` user)
6. Select a measure and scroll down to Definition of Population and make sure that all the stuff under Definition of denominator matches the appropriate new definitions in the description of [this](https://jiraent.cms.gov/browse/CMDCT-3689) ticket

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -19,88 +20,182 @@ interface Props {
...
+      },
+    ],
+    description:
+      "Medicaid (Title XIX) beneficiaries, Medicaid-Expansion CHIP (Title XXI) beneficiaries,",
+  },
+  CCSM: {
+    options: [
+      {
+        displayValue: "Medicaid (Title XIX)",
+        value: DC.DENOMINATOR_INC_MEDICAID_POP,
```
angelaco11, line 49: Not sure about these values. What are they used for? Do child measures need have different values than standard measures? would love any amount of context you all have here

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -19,88 +20,182 @@ interface Props {
...
   }
 
   return (
     <CUI.Box>
       <CUI.Text mt="3">
-        {`Please select all populations that are included. For example, if your data include both non-dual Medicaid ${
-          healthHomeMeasure ? "enrollees" : "beneficiaries"
-        } and Medicare and Medicaid Dual Eligibles, select both:`}
+        {`Please select all populations that are included in the denominator. For example, if your data include 
+        ${coreSetOptions[coreSetType].description}
```
angelaco11, line 126: I know this seems kind of weird to have partial sentences stored as constants for each core set. But this seemed the cleanest to me instead of doing aggressive amounts of conditional string rendering inside this Text element (they're different for 3 core sets, medicaid, CHIP and health home). I also didn't want to store the whole description (starting with `For example`) in the coreSetOptions descriptions since it was redundant and would clutter that object even more. 

Do you guys have any opinions here? I'm open to any suggestions

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -19,88 +20,182 @@ interface Props {
...
+        displayValue: "Medicaid (Title XIX)",
+        value: DC.DENOMINATOR_INC_MEDICAID_POP,
+      },
+      {
+        displayValue: "Medicaid-Expansion CHIP (Title XXI)",
+        value: DC.DENOMINATOR_INC_MEDICAID_EXPANSION,
+      },
+    ],
+    description:
+      "Medicaid (Title XIX) beneficiaries, Medicaid-Expansion CHIP (Title XXI) beneficiaries,",
```
angelaco11, line 43: see [this](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2279/files#r1635177499) comment for the reasoning behind these partial descriptions

ailZhou (changes requested):
At a quick glance, the changes are appearing locally for me and it looks to be matching the ask. i also appreciate the organization you've added to this file, its definitely way more readable of what the options labels are for each coreSet 🎉 

unfortunately i need to ask for a change. so the "fun" thing about QMR updates, all changes are only meant for the current year and forward, previous years structure has to be preserved like its in a museum. This component is shared for 2021 - 2023 reports as well, so we need keep those labels along with these new labels in this file somehow. Or if you think its different enough and a new file should be made?

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -172,6 +390,10 @@ export const DefinitionOfPopulation = ({
   //WIP: using form context to get the labels for this component temporarily.
   const labels: any = useContext(SharedContext);
 
+  const params = useParams();
+  const coreSet = params.coreSetId;
+  const yearIs2024 = location.pathname.includes("/2024");
```
angelaco11, line 395: let me know if there's a better way to detect the year in these shared components. I know the parent components that renders DefintionOfPopulation has the year, but then I would have to pass it through as a prop approximately 10,000 times lol

angelaco11:
> At a quick glance, the changes are appearing locally for me and it looks to be matching the ask. i also appreciate the organization you've added to this file, its definitely way more readable of what the options labels are for each coreSet 🎉
> 
> unfortunately i need to ask for a change. so the "fun" thing about QMR updates, all changes are only meant for the current year and forward, previous years structure has to be preserved like its in a museum. This component is shared for 2021 - 2023 reports as well, so we need keep those labels along with these new labels in this file somehow. Or if you think its different enough and a new file should be made?

This is a great callout, thank you Aileen! I have officially updated my PR to address this change request

ailZhou (dismissed):

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -11,20 +11,27 @@ import * as DC from "dataConstants";
...
   hybridMeasure?: boolean;
   populationSampleSize?: boolean;
   healthHomeMeasure?: boolean;
 }
+interface DefOfDenomOption {
+  displayValue: string;
+  value: string;
+}
+interface CoreSetSpecificOptions {
+  [id: string]: {
```
rocio-desantiago, line 27: is this `id` the same as `coreSetId`?

rocio-desantiago (dismissed):
Thank you for the context Angela 👍

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
angelaco11: I updated these names to be more specific! They're slightly different so I named each one accordingly

benmartin-coforma (dismissed):

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx.rej
benmartin-coforma, line 1: We'd better leave this file out; I think it was auto-generated by `git apply`

angelaco11: ooh good catch!!!

benmartin-coforma (approved):
I am a fan of the code here, and I'd like to get this merged in sooner than later. The fact that it hasn't deployed to an ephemeral environment means that I can't run any tests there to give myself warm fuzzy feelings.

Maybe the actions are misbehaving because of the fork? Or some other one-off weirdness? This may be a case where it makes sense to merge into master first, in order to run our deployed-env tests. Especially given I don't see how anything in this PR could behave differently deployed vs locally.

BearHanded (approved):

PR #2282, opened 2024-06-13
remove unit test branch lock

gmrabian:
### Description

Our unit tests workflow was only running on `master`

Now it will run in feature branches, val, and prod as well.

It runs alongside the deploy command so there is no time spent waiting for unit tests to finish before deploying

See [this branch's deploy action](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/9500897568) as an example

### Related ticket(s)

CMDCT-3575
CMDCT-3576

---
### How to test

See [this branch's deploy action](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/actions/runs/9500897568) as an example

---
### Pre-review checklist

### DIFF
--- a/.github/workflows/unittest-workflow.yml
+++ b/.github/workflows/unittest-workflow.yml
@@ -23,8 +23,7 @@ jobs:
         with:
           path: |
             **/node_modules
-            ~/.cache/Cypress
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', 'plugins/**') }}
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
       - name: set path
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/unittest-workflow.yml
```diff
@@ -23,8 +23,7 @@ jobs:
         with:
           path: |
             **/node_modules
-            ~/.cache/Cypress
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', 'plugins/**') }}
```
gmrabian, line 27: Cypress not needed for unit tests
plugins/ folder no longer exists

rocio-desantiago (approved):
Got come context about the runners running in parallel 👍 
Knowing that `the unit tests run in parallel to the deploy step` gives me peace that this change shouldn't slow anything down with GH actions

PR #2311, opened 2024-07-02
QMR - IET-AD validation for totals matching sum

rocio-desantiago:
### Description

Spanking brand new validation to account for validating the sums vs pre-generated totals at the the bottom of the `IET-AD` measure page. 

There are 12 inputs for each `category` (3 SUD treatment types plus their `label` - `Ages 18 to 64` vs `Age 65 and older` for `Initation of SUD Treatment` vs `Engagement of SUD Treatment` for each of these Treatment Types )
1. Alcohol Use Disorder
   a. Ages 18 to 64
   b. Age 65 and older
2. Opioid Use Disorder
   a. Ages 18 to 64
   b. Age 65 and older
3. Other Substance Use Disorder
   a. Ages 18 to 64
   b. Age 65 and older

Numerators and Denominators from these categories and respective age groups are added and a total is auto-generated, but the user can still change these totals, this validation produces a generic error if a total is changed and does not match the sum so that user is aware.

### Related ticket(s)

CMDCT-3567

---
### How to test

1. Enter numbers for at least 2 treatment categories and at least one age group. (Blank ones will be ignored)
2. Update the total value to a different number to confirm that the error shows up upon clicking `Validate Measure`
   a. The error should say "Numerators must be the sum for each category of performance measures" if Numerator total is manually changed.
   b. The error should say "Denominators must be the sum for each category of performance measures" if Denominator total is manually changed. **NOTE:** Both denominators for initiation or engagement must be changed and match as that produces a different error on top of the new error from this validation.

### Screenshot
In a theoretical world if all totals were updated for both age groups, up to 8 errors are possible to show up (although not likely). 

Numerator error
Error for numerator when total does not match sum

Denominators must match for each age group
Total values for denominators match for each group

Denominator error shows up twice due to matching
Error for denominator when total does not match sum

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/measures/2024/shared/globalValidations/validateNDRTotalsMatchSum/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  it("should return numerator sum error - sums do not match totals for initiation values", () => {
+    const valNumerator = validateNDRTotalsMatchSum(pmaNumeratorInitiationSum);
+
+    expect(valNumerator.length).toBe(1);
+    for (const result of [...valNumerator]) {
+      expect(result.errorLocation).toBe("Performance Measure");
+      expect(result.errorMessage).toBe(
+        "Numerators must be the sum for each category of performance measures"
+      );
+    }
```
benmartin-coforma, line 25: Suggestion: We can simplify this a bit.

The `[...foo]` syntax is a [rest operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#rest_property). It's useful if you want to
* Combine multiple arrays into one: `combined = [...arrA, ...arrB]`
  - Note the alternative syntax of `combined = arrA.concat(arrB)`
* Turn something array-ish, like a NodeList, into a bona fide array: `nodeList = document.querySelectorAll("p"); nodeArray = [...nodeList]`
  - Note the alternative syntax of `nodeArray = Array.from(nodeList)`
* Make a shallow copy of an array: `arrA = [...arrB]; console.log(arrA === arrB) /*false*/; console.log(arrA[0] === arrB[0]) /*true*/`
  - Note that `Array.from` still does the job here, as well as tricks like `arrA = arrB.slice()`

We're not doing any of those things, so we can just loop `for (const result of valNumerator) { ... }`.

But we can go a step further! We've just validated that `valNumerator` has length exactly 1, so we don't really need the loop at all.

```suggestion
    const = valNumerator[0];
    expect(result.errorLocation).toBe("Performance Measure");
    expect(result.errorMessage).toBe(
      "Numerators must be the sum for each category of performance measures"
    );
```

And this applies to all three tests below, as well 👍

benmartin-coforma (approved):
Very well organized! Nice work.

services/ui-src/src/measures/2024/shared/globalValidations/validateNDRTotalsMatchSum/index.test.ts
rocio-desantiago: @benmartin-coforma Thank you for noting that!

benmartin-coforma (approved):

ailZhou (approved):
the code looks good! i think there might be a bug in the rate cal but that doesn't look to be affecting your code so feel free to merge in.

rocio-desantiago:
> the code looks good! i think there might be a bug in the rate cal but that doesn't look to be affecting your code so feel free to merge in.

Leaving open as discussed.
Team to merge upon design review.

PR #2314, opened 2024-07-03
Inconsistent Rounding

ailZhou:
### Description

Using .toFixed() for rounding seems to cause a weird issue were it only rounds half the time, so I'm switching to good ol' shift forward by two decimals, round and then shift back by two decimals. We've all seen and hate this, if anyone knows a better way, please let me know!

[This explains the issue well](https://stackoverflow.com/questions/77144696/seemingly-inconsistent-tofixed-behavior). Scroll to the **toFixed(1) example** example

### Related ticket(s)

CMDCT-3780

---
### How to test

1) Sign into QMR, any separate user. i.e. [stateuser1@test.com](mailto:stateuser1@test.com)
2) Enter ADD-CH: Medicaid
3) Fill out Measure Specification to trigger the Performance Measure layer
4) For the Data Source: Select Administrative Data.
5) In the Performance Measure Section:
   - Enter Numerator: 305 
   - Enter Denominator: 10000
   - The Rate should generate to 3.1 (actual rate would be 3.05 but rounded to 3.1)
6) Save
7) Do steps 3 - 5 for ADD-CH: CHIP
8) Go to the Combined Rates Page for this Measure. It should look like this
Screenshot 2024-07-09 at 4 44 09 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/utils/constants/math.ts
+++ b/services/app-api/utils/constants/math.ts
@@ -1,0 +1,7 @@
+//this function is copied from in ui-src/utils/rateFormulas/index.ts, we should merge this into a shared file if we ever figure it out
+export const fixRounding = (value: number, numbersAfterDecimal: number) => {
+  return (
+    Math.round((value + Number.EPSILON) * Math.pow(10, numbersAfterDecimal)) /
+    Math.pow(10, numbersAfterDecimal)
+  );
+};
--- a/services/app-api/handlers/rate/calculations/rateCalculation.ts
+++ b/services/app-api/handlers/rate/calculations/rateCalculation.ts
@@ -1,4 +1,5 @@
 import { Program, StandardRateShape } from "../../../types";
+import { fixRounding } from "../../../utils/constants/math";
 import { DataSource, FormattedMeasureData } from "./types";
 
 export abstract class RateCalculation {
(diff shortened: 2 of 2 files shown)


BearHanded (approved):
Works great in the deployed environment.

benmartin-coforma (approved):
I'd like to know what the original author intended the `+ Number.EPSILON` to accomplish, because I'm not convinced it needs to be there. But, for the purposes of this PR, we'd better leave it alone.

Approved!

services/app-api/utils/constants/math.ts
```diff
@@ -0,0 +1,7 @@
+//this function is copied from in ui-src/utils/rateFormulas/index.ts, we should merge this into a shared file if we ever figure it out
+export const fixRounding = (value: number, numbersAfterDecimal: number) => {
+  return (
+    Math.round((value + Number.EPSILON) * Math.pow(10, numbersAfterDecimal)) /
```
rocio-desantiago, line 4: Oooh `Number.EPSILON` - first time seeing this one. I see that it "represents the difference between 1 and the smallest floating point number greater than 1." - per MDN web docs. 

And it's value is `2.220446049250313e-16`. Clarifying question, is the purpose of adding Number.EPSILON to round to the nearest decimal point for accuracy?

rocio-desantiago (commented):
Question about Number.EPSILON

services/app-api/utils/constants/math.ts
ailZhou: not in this instance, its the .toFixed(1) that does the rounding. this is basically a copy over of what we already have that the client approved of for the rate cals in the front end

rocio-desantiago: Gotcha 👍

PR #2315, opened 2024-07-08
Mandatory Measure Instructions - Home Health

ajaitasaini:
### Related ticket(s)

CMDCT-3775

---
### How to test

1. Add a HH core set page for FFY 2024 and confirm that the banner exists. Banner should match [Figma design](https://www.figma.com/design/CaKupDWdVLgmnFZyhBSdGm/MDCT_QMR?node-id=313-2339&t=nqgSCIgA14HNsDl1-0). Info text should be: **"States with approved Health Home Programs in operation by June 30, 2023 are required to report all of the measures on the Health Home Core Set. More information on mandatory reporting requirements is included in the Initial Core Set Mandatory Reporting Guidance for the Health Home Core Sets."**
2. Confirm that the banner does not exist prior to 2024

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/views/CoreSet/index.tsx
+++ b/services/ui-src/src/views/CoreSet/index.tsx
@@ -313,6 +313,8 @@ export const CoreSet = () => {
     (measure) => measure.id
   );
 
+  const coreSetPrefix = coreSet[0].slice(0, 2);
+
   return (
     <QMR.StateLayout
       breadcrumbItems={[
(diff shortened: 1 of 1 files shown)


services/ui-src/src/views/CoreSet/index.tsx
```diff
@@ -323,17 +325,18 @@ export const CoreSet = () => {
...
-            </CUI.Text>
-          </Alert>
-        </CUI.Box>
-      )}
+      {Number(year) >= 2024 &&
+        (coreSetPrefix === "CC" || coreSetPrefix === "HH") && (
+          <CUI.Box mb="8">
+            <Alert heading="Mandatory Measure Instructions">
+              <CUI.Text>
+                {coreSetPrefix === "CC"
```
BearHanded, line 333: I think it is worth reducing complexity by only switching on this value once

rocio-desantiago: Just for context: 
I'm trying to follow the boolean here. 
Is one of the statements an if else statement checking for certain text? So as long as either of those paragraphs exists, this is true?

rocio-desantiago (commented):
Question about the boolean statement

BearHanded (approved):

rocio-desantiago (approved):

services/ui-src/src/views/CoreSet/index.tsx
ajaitasaini: @rocio-desantiago the original statement was splicing the core set string to determine what kind of core set we're looking at. Using that spliced string, we confirmed if it was a child or health home core set, and if so we'd display a warning. If it was a child core set, we display the warning with the first paragraph string, and if it was a health home core set, we display the warning with the second paragraph string. With Brendan's help, I changed the code so there's one less check -- hopefully it's a little easier to read!

PR #2316, opened 2024-07-09
CMDCT-3472: adding some nightmarish manual brute force styles to pdfs

angelaco11:
### Description
This actually ended up being a pretty contained PR so I am not going to say sorry for it even though it is truly an abomination. 

### **Summary of Prince issues in our PDFs**

Prince PDFs have a long and well known limitation that it does not recognize certain css properties like `background-color` (see [this](https://stackoverflow.com/questions/20653871/background-color-is-not-showing-in-prince-xml) stack overflow post describing the issue and solution). I understand this stack overflow post is 10 years old, but our PDFs were having the EXACT issues described in that post with certain chakra elements like radio buttons and checkboxes. Someone mentioned that you can override these styles using !important, so that's what this PR is. Any styles that aren't properly coming through are recreated in the custom prince styling that we are already doing. 

Also Prince PDFs do not support css grid, so I did a little manual overriding there too.

Also also, for some god forsaken reason, we were already going through every bit of css and changing any instance of `display: flex `to `block` and then `text-align: right` to `center` (see [here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/services/ui-src/src/views/ExportAll/util.tsx#L62) for the chunk of code that does this). That works for MOST elements but it was messing up some elements too. I found it easiest to find the specific ones that were broken after doing that and manually override them to have the correct css again. Not ideal at all but it does sort of work!

### **Summary of the things I fixed**

1. Chakra buttons did not have a grey background and they also were not displaying with proper spacing because they are in a css grid which prince pdfs doesn't support. I just manually added the background and recreated the grid spacing with margins. See before and after: 
Screenshot 2024-07-09 at 9 57 04 AM

2. Chakra radio buttons and checkboxes only had an outline (background color was not taking affect) so I fixed that. See below: 
Screenshot 2024-07-08 at 4 00 39 PM

3. In the questionnaire section, there were some percentages that were all messed up because of the thing i referenced about us replacing `display: flex` with `display: block`. These percentages need `display: flex` to look normal so I just manually overrode (overWrote?) that css to make it look gud:
Screenshot 2024-07-09 at 10 37 33 AM

4. In the sections where there are Numerator, Denominator, and Rate inputs, `display: flex` was not getting applied correctly and it was being displayed as a column instead of row. so I added flex back in so that the display is a row: 
Screenshot 2024-07-09 at 11 22 53 AM

### Related ticket(s)
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2142

---
### How to test
1. Go to QMR core set view and click on the three dot menu of any core set (ideally core sets where you've completed some measures) and click "Export"
2. This will take you to the pdf print out page. Click the "Print PDF" button and wait for the PDF to export
3. Once the PDF opens in a new tab, give it a good look-see. Make sure it sort of looks as good as possible and use my screenshots as a reference to make sure that all my claimed changes are actually showing up

### DIFF
--- a/services/ui-src/src/views/ExportAll/util.tsx
+++ b/services/ui-src/src/views/ExportAll/util.tsx
@@ -123,25 +123,39 @@ export const applyPrinceSpecificCss = (): HTMLStyleElement => {
     .prince-top-link, .prince-supp-text, h1 { margin: auto !important; text-align: center !important; width: fitcontent !important; margin: 10px 0 !important; }
     .prince-upload-wrapper, .prince-file-item { border: 3px !important; border-style: dotted; background-color: var(--chakra-colors-blue-100); border-radius: var(--chakra-radii-md) }
     .replaced-text-area {border-radius: var(--chakra-radii-md); border-width: 1px; border-style: solid; border-color: inherit; padding: 15px; box-sizing: border-box; white-space: pre-wrap;}
+    
+    
     ${
-      /* Prince doesn't support color and background colors or css grid out of the box. we have to brute force those styles here */ ""
+      /* ******* Prince doesn't support certain css elements like background colors or css grid out of the box. we have to brute force those styles here.
+      All of these css class names can be found by inspecting the elements on the pdf page in the browser. ********* */ ""
     }
-    .chakra-button { background: var(--chakra-colors-gray-100) !important; margin: 16px }
+
+    .chakra-button { background: var(--chakra-colors-gray-100) !important; margin: 16px 8px }
     .chakra-link { color: var(--chakra-colors-blue-600) !important; }
     .chakra-link * { color: blue !important; }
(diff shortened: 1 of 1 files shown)


benmartin-coforma:
This is incredible work, thank you so much.

And I hate to pile on but I did notice this difference in the webpage vs the PDF - do you think it could be solved with a similar technique?

HTML:
webpage render

PDF
PDF render

angelaco11:
> This is incredible work, thank you so much.
> 
> And I hate to pile on but I did notice this difference in the webpage vs the PDF - do you think it could be solved with a similar technique?


This has been fixed, thank you Ben for your thorough testing that caught this!!

BearHanded (approved):
Checked in the deployed env.

This is still not pixel perfect, but as far as the spike goes, miles better than where we started with a real understanding of the problem and a path forward if pixel perfect is ever required.

Solid work.

ailZhou (approved):
This looks amazing!!! Thank you for fixing so much of the mangled display in the export 🎉 !

PR #2318, opened 2024-07-09
CMDCT-3784, CMDCT-3786, CMDCT-3789: updating labels for combined rates and other things

angelaco11:
### Description
This super simple PR contains label changes for 3 tickets:
1. CMDCT-3784: Changed "CHIP" to "Separate CHIP"
2. CMDCT-3786: Changed "Not answered" to "Not reported"
3. CMDCT-3789: Changed 2024 FUA-AD measure to capitalize the up in "Follow-Up"

---
## Testing CMDCT-3784 and CMDCT-3786
1. Navigate to combined rates page and click into any measure for child or adult
2. Observe that any reference to CHIP now says Separate CHIP
3. Observe that anything that doesn't have a value says "Not reported" instead of "Not answered"

## Testing CMDCT-3789:
1. Navigate to core sets and click on Adult Medicaid
2. Scroll down to the FUA-AD measure and make sure the title has "up" capitalized in "Follow-Up"
3. Click into the measure and make sure that the title and the first question all have "Follow-Up" capitalized correctly

### DIFF
--- a/services/ui-src/src/measures/measureDescriptions.ts
+++ b/services/ui-src/src/measures/measureDescriptions.ts
@@ -18,7 +18,7 @@ export const measureDescriptions: MeasureList = {
     "CPA-AD":
       "Consumer Assessment of Healthcare Providers and Systems (CAHPS®) Health Plan Survey 5.1H, Adult Version (Medicaid)",
     "FUA-AD":
-      "Follow-up After Emergency Department Visit for Alcohol and Other Drug Abuse or Dependence",
+      "Follow-Up After Emergency Department Visit for Alcohol and Other Drug Abuse or Dependence",
     "FUH-AD":
       "Follow-Up After Hospitalization for Mental Illness: Age 18 and Older",
     "FUM-AD": "Follow-Up After Emergency Department Visit for Mental Illness",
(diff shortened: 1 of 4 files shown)


services/ui-src/src/measures/measureDescriptions.ts
```diff
@@ -319,7 +319,7 @@ export const measureDescriptions: MeasureList = {
     "CPU-AD":
       "Long-Term Services and Supports Comprehensive Care Plan and Update",
     "FUA-AD":
-      "Follow-up After Emergency Department Visit for Substance Use: Age 18 and Older",
+      "Follow-Up After Emergency Department Visit for Substance Use: Age 18 and Older",
```
angelaco11, line 322: I know [this](https://jiraent.cms.gov/browse/CMDCT-3789) ticket is under the 2024 updates epic, but this feels like a typo that should be updated for all years because all the other measures in the other years have Follow-Up. Anyone have any opinions? I can change it to only update for 2024 but i won't like it

BearHanded: I think this should be pretty innocuous but just let David know all the same. The folks that test this in val really are hyper-sensitive to changes.

BearHanded (approved):

services/ui-src/src/measures/measureDescriptions.ts
angelaco11: Checked with David and this change is only for 2024 so I updated accordingly. Good call on checking on that Brendan!

BearHanded (approved):

ailZhou (approved):
lgtm! tested on deploy and the labels are updated

PR #2319, opened 2024-07-10
Combined Rates Page - Rates showing in wrong order

ailZhou:
### Description

The BO's want the table in Combined Rates to display the labels of the N/D/R set the same way it is listed in the Performance Measure section. This PR will address that.

How this works is we grab the categories and qualifies from the rateLabelText of whatever year the user is looking at by doing a semi-dynamic import:
Screenshot 2024-07-10 at 4 14 34 PM

Then we build an ordered array containing the uid and do a sort to reorganize the table to match the N/D/R order in the PM section.
Screenshot 2024-07-10 at 2 50 03 PM

How the data looks in the Measure (example AMM-AD):
Screenshot 2024-07-10 at 2 56 39 PM

Currently what it's doing:
Screenshot 2024-07-10 at 2 56 19 PM

What it should look like:
Screenshot 2024-07-10 at 2 56 59 PM

### Related ticket(s)

CMDCT-3782

---
### How to test

Deploy Link: https://d1bhfisusdxz9x.cloudfront.net/

1) Sign into QMR, any user with combined rates. i.e. [stateuser@test.com](mailto:stateuser@test.com)
2) Fill in N/D/R information for any of these measures: AMM-AD, FUM-AD, IET-AD, APM-CH, TFL-CH
3) Save your data
4) Go to the Combined Rates page and check that the combine rates - N/D/R table is the same order wise as the measure's page.

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
+++ b/services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
@@ -4,7 +4,6 @@ import {
   RateCategoryMap,
   RateDataShape,
 } from "./CombinedRateTypes";
-import { useEffect, useState } from "react";
 import { LabelData } from "utils";
 import * as Labels from "labels/RateTextLabels";
 
@@ -86,23 +85,18 @@ export const sortRateNDR = (
 };
 
 export const CombinedRateNDR = ({ json }: Props) => {
-  const [tables, setTables] = useState<TableDataShape[]>();
-  const measure = json?.measure!;
-  const year = json?.year!;
+  const measure = json.measure;
+  const year = json.year;
   const data = collectRatesForDisplay(json);
   const headers: ProgramType[] = ["Medicaid", "Separate CHIP", "Combined Rate"];
   const rows: Measures[] = ["numerator", "denominator", "rate"];
 
   //dynamically pull the rateLabelText by combined rates year so that we can get the cat and qual info of the measure
   const rateTextLabel = Labels[`RateLabel${year}` as keyof typeof Labels];
-  useEffect(() => {
-    if (rateTextLabel) {
-      const { categories, qualifiers } = rateTextLabel!.getCatQualLabels(
-        measure! as keyof typeof rateTextLabel.data
-      );
-      setTables(sortRateNDR(data, categories, qualifiers));
-    }
-  }, [rateTextLabel]);
+  const { categories, qualifiers } = rateTextLabel!.getCatQualLabels(
+    measure! as keyof typeof rateTextLabel.data
+  );
+  const tables = sortRateNDR(data, categories, qualifiers);
 
   //centralize formatting of the display data so that all the renders value are consistent
   tables?.forEach((table) => {
@@ -154,11 +148,8 @@ type TableDataShape = {
 };
 
 const collectRatesForDisplay = (
-  json: CombinedRatePayload | undefined
+  json: CombinedRatePayload
 ): TableDataShape[] => {
-  if (!json) {
-    return [];
-  }
   const tables: TableDataShape[] = [];
   const data = json.data;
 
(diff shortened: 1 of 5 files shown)


BearHanded (approved):
We just need to remember to update the readme when we do a pass of it this year with the new RateTextLabel file.

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
```diff
@@ -70,13 +73,39 @@ const HorizontalTable = (
...
+  //dynamically pull the rateLabelText by combined rates year so that we can get the cat and qual info of the measure
+  const rateTextLabel = Labels[`RateLabel${year}` as keyof typeof Labels];
+  useEffect(() => {
+    if (rateTextLabel) {
+      const { categories, qualifiers } = rateTextLabel!.getCatQualLabels(
+        measure! as keyof typeof rateTextLabel.data
+      );
+      setTables(sortRateNDR(data, categories, qualifiers));
+    }
+  }, [rateTextLabel]);
```
benmartin-coforma, line 105: Question: I don't understand the purpose of UseEffect here. Isn't everything happening synchronously here (once we have the input `json`)?

ailZhou: I dunno if its me, but there's a slight delay when running `const rateTextLabel = Labels[`RateLabel${year}` as keyof typeof Labels];` and it returns an undefined at first. So I needed a useEffect because of that

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
```diff
@@ -70,13 +73,39 @@ const HorizontalTable = (
...
 
+  //dynamically pull the rateLabelText by combined rates year so that we can get the cat and qual info of the measure
+  const rateTextLabel = Labels[`RateLabel${year}` as keyof typeof Labels];
+  useEffect(() => {
+    if (rateTextLabel) {
+      const { categories, qualifiers } = rateTextLabel!.getCatQualLabels(
+        measure! as keyof typeof rateTextLabel.data
+      );
+      setTables(sortRateNDR(data, categories, qualifiers));
+    }
```
rocio-desantiago, line 104: Question: Is this 'useEffect' function run more than one time?

rocio-desantiago (commented):
Small question

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
ailZhou: useEffect runs when there's a change to the observe data in the [] near the end of the call, so for this one it will run whenever it sees that the variable [rateTextLabel] has been updated.

BearHanded (approved):

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
benmartin-coforma: Ah, I see what's happening. When the component first renders, the combined rate data hasn't loaded at all. Therefore `year` is undefined, therefore `Labels["RateLabelundefined"]` is undefined, etc etc. This is hidden by the `!` operators earlier in the component.

I think we could solve this by skipping more code when `data` is undefined - whereas currently we continue down the same code path with `!` and `?` as required to appease the compiler. That could be simpler overall, but it would also be a change of convention.

I could apply that refactor if you like, either as a new commit on top of this PR, or in a new PR later. But I'm also fine with the code as-is, so it's up to you!

benmartin-coforma (approved):

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
ailZhou: you can just add it to this PR, i don't mind. I trust that whatever you write is a benefit to the app 👍.

benmartin-coforma (approved):
Looks awesome ;)

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
benmartin-coforma: This `useEffect` is no longer in the component. Conversation resolved! :D

BearHanded (approved):

PR #2320, opened 2024-07-11
Combined Rates Page - Strict check & Not Reporting Behavior

ailZhou:
### Description

**Issue no. 1**
The current Combined Rates calculation was doing a very loose matching of data source. It check that as long as the value existed in the source, it would calculate. I changed the the `check` function in **adminstrativeCalculation.ts** to be a strict match. i.e.

**This would fail**
Medicaid: ["Administrative Data"]
CHIP: ["EHR", "Administrative Data"],

**This would pass**
CHIP: ["Administrative Data"]
Medicaid: ["EHR", "Administrative Data"],

**Issue no. 2**
We want to run the calculation if either a CHIP or Medicaid measure was filled out and not wait until both have been filled. I added a new function `validateMeasureData`, it checks each data to see if its valid and and adds them to an array that we return. That valid data is stored in **validatedData** and we run the calculations again this data. 

### Related ticket(s)

CMDCT-3770
CMDCT-3787

---
### How to test

Deploy Link: https://d2wus07wq3kx87.cloudfront.net/

1. Sign into QMR, any user with combined rates. i.e. [stateuser@test.com](mailto:stateuser@test.com)
2. Enter ADD-CH: Medicaid
    - Fill out Measure Specification to trigger the Performance Measure layer
    - For the Data Source: Select Administrative Data.
    - Fill out the Performance Measure Section
    - Save
7. Go to the combined rates page for ADD-CH and see that the rate is calculated even with one set of data
8. Enter ADD-CH: CHIP
     - In the **Are you reporting on this measure?** Section, select No and save
9. Check the Combined Rates page to see that a Combined Rate still exist 
10. Go back to ADD-CH: CHIP
    - For the Data Source: Select Administrative Data & Other Data Source.
    - Save
11. Check the Combined Rates page and no the Combined Rates should not have calculated.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


services/app-api/handlers/rate/shared/calculateAndPutRate.ts
```diff
@@ -19,6 +19,17 @@ export const formatMeasureData = (data: (Types.Measure | undefined)[]) => {
...
+export const validateMeasureData = (data: FormattedMeasureData[]) => {
+  const validData: FormattedMeasureData[] = [];
+  data.forEach((program) => {
+    if (Object.values(program.rates).length > 0) {
+      validData.push(program);
+    }
+  });
+  return validData;
+};
+
```
benmartin-coforma, line 32: Suggestion: This is the classic use case for Array..filter:

```suggestion
// Checks to see which set of data is valid, and only return those for calculation
export const removeEmptyRates = (data: FormattedMeasureData[]) => {
  return data.filter((program) => Object.values(program.rates).length > 0);
};

```

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
```diff
@@ -37,17 +48,20 @@ export const calculateAndPutRate = async (
     const data = await getMeasureByCoreSet(combinedCoreSet!, pathParameters);
     const formattedData = formatMeasureData(data);
 
+    //check which data is valid for summation
+    const validatedData = validateMeasureData(formattedData);
```
benmartin-coforma, line 52: Suggestion: to match the function rename above

```suggestion
    const validatedData = removeEmptyRates(formattedData);
```

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
benmartin-coforma, line 1: I have a few suggestions here.
1. Let's de-nest with an early return for hybrid data sources
2. Since we don't want to perform the calculation if both CHIP and Medicaid have empty data source arrays, and we know that before we enter the loop, we can skip the loop and early return for that too
3. If, within the loop, we know that at least one measure has data source(s), we can skip the length checks entirely
4. This is a great time to use `Array..some()`

All that adds up to this version of the function:
```ts
  check(arr: FormattedMeasureData[]): boolean {
    const chipSources =
      arr.find((data) => data.column === "CHIP")?.dataSource ?? [];
    const medicaidSources =
      arr.find((data) => data.column === "Medicaid")?.dataSource ?? [];

    // If neither measure has any data source, we will not use this calculation
    if (chipSources.length === 0 && medicaidSources.length === 0) {
      return false;
    }

    // If the user had selected hybrid as a data source, we will not use this calculation
    const isHybrid = [chipSources, medicaidSources].some((srcs) => {
      return (srcs as string[])?.includes(DataSource.Hybrid);
    });

    if (isHybrid) {
      return false;
    }

    return this.dataSrcMap.some((dataSrc) => {
      // If a measure has no data sources, .every() returns true
      // This is good because we still want to calculate even if only one measure is complete.
      const chipSourcesMatch = chipSources.every((chipSrc) => dataSrc.CHIP.includes(chipSrc));
      const medicaidSourcesMatch = medicaidSources.every((medSrc) => dataSrc.Medicaid.includes(medSrc));
      return chipSourcesMatch && medicaidSourcesMatch;
    });
  }
```

benmartin-coforma (commented):
Makes sense to me! Comments below.

services/app-api/handlers/rate/shared/calculateAndPutRate.ts
ailZhou: all hail the almighty filter! great catch, thanks!

services/app-api/handlers/rate/calculations/adminstrativeCalculation.ts
ailZhou: Thank you for breaking it down and suggesting how to get rid of the nested if statements! This is so much more readable 🎉

BearHanded (approved):

benmartin-coforma (approved):

ajaitasaini (approved):

PR #2324, opened 2024-07-12
CMDCT-3790: Fixing a few different bugs in involving validation for other measurement specification rate calcs

angelaco11:
### Description
There was an issue where for certain Health Home measures (AIF-HH and IU-HH to be specific) if you selected "Other" for Measurement Specification, filled out the form and filled out some of the Optional Measure Stratification (OMS) rates, the measure would save successfully, but it would fail to complete with an error in the console:
Screenshot 2024-07-12 at 1 29 37 PM

The cause of this bug is that during the validation process, we were attempting to iterate on an object and the object we were trying to iterate over never even existed in the first place. 

I just removed the offending object iteration because it wasn't really being used anyway. I then changed the way that we grab the optional measure stratification calculations to be validated. This fixed the console error and now the measure can be completed successfully and validates properly. 

---
### How to test

1. Sign into a state that has Health Home coreset ([stateuserNC@test.com](mailto:stateuserNC@test.com) or [stateuserDC@test.com](mailto:stateuserDC@test.com) are two that I know of)
5. Click on the Health Home coreset and navigate to either the AIF-HH or the IU-HH measure
6. Fill out the form with "Other" selected for Measurement Specification
7. When you get to the Optional Measure Stratification, fill out at least one of the fields for Race, Ethnicity, etc.
8. Validate the measure and make sure there is no console error like the one in the description
9. Complete the measure and make sure that the measure truly completes without the console error in the description
10. do some additional testing to make sure that validation for the optional measure stratification rates still works as expected
11. repeat steps 1-8 for 2023 (i'm pretty sure only DC user has HH for '23)


services/ui-src/src/measures/2021/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
```diff
@@ -6,21 +6,22 @@ interface NDRforumla {
```
angelaco11, line 6: This file fixes a validation bug specifically in the '21 and '22 measurements ('22 file fix is [here](services/ui-src/src/measures/2022/shared/globalValidations/omsValidator/index.ts)) where it would claim that optional measurement stratification was not filled out correctly when it was indeed filled out correctly. 

This is not NECESSARILY related to the bug at hand, but it was unearthed once I fixed the bug so I included a fix in this PR just so that I could feel satisfied that the whole feature actually works now.

services/ui-src/src/measures/2021/AIFHH/validation.ts
```diff
@@ -100,14 +100,10 @@ const OMSValidations: GV.Types.OmsValidationCallback = ({
   locationDictionary,
   label,
 }) => {
-  const rates = Object.keys(rateData?.rates ?? {}).map((x) => {
-    return { rate: [rateData?.rates?.[x]?.OPM?.[0]] };
```
angelaco11, line 104: This object never existed so this was always returning `{rate: undefined}`. I got rid of this because it didn't really make sense to me why we were doing this in the first place when we are already passing the rateData through to the next function and that has everything we need.

services/ui-src/src/measures/2021/globalValidations/omsValidator/index.ts
angelaco11, line 1: This file fixes a validation bug specifically in the '21 and '22 measurements ('22 file fix is [here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2324/services/ui-src/src/measures/2022/shared/globalValidations/omsValidator/index.ts)) where it would claim that optional measurement stratification was not filled out correctly when it was indeed filled out correctly.

This is not NECESSARILY related to the bug at hand, but it was unearthed once I fixed the bug so I included a fix in this PR just so that I could feel satisfied that the whole feature actually works now.

services/ui-src/src/measures/2021/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
```diff
@@ -6,21 +6,22 @@ interface NDRforumla {
...
   ndrFormulas: NDRforumla[],
   errorLocation: string
 ) => {
   let errorArray: any[] = [];
   for (const key in rateData) {
-    if (OPM && OPM.length > 0) {
-      errorArray.push(
-        ...ComplexNoNonZeroNumOrDenom(
-          [],
-          [{ description: key, rate: [...rateData[key]["OPM"]] }],
```
angelaco11, line 19: this was the main source of the bug. `[...rateData[key]["OPM"]]` does not exist so it was throwing the console error that i screenshotted in the description. I updated this to actually grab the correct rate and pass it through.

benmartin-coforma (dismissed):
I have not tested, but this looks good to me 👍

services/ui-src/src/measures/2023/AIFHH/validation.ts
```diff
@@ -100,22 +100,17 @@ const OMSValidations: GV.Types.OmsValidationCallback = ({
   locationDictionary,
   label,
 }) => {
-  const rates = Object.keys(rateData?.rates ?? {}).map((x) => {
-    return { rate: [rateData?.rates?.[x]?.OPM?.[0]] };
```
angelaco11, line 104: The `rateData?.rates?.[x]?.OPM` is the OPM legacy data structure (for years '21 and '22). '23 and '24 measures OPM data structure has been updated to be `rateData?.rates?.OPM?.[x]` so this function is outdated and was causing bugs because that structure doesn't exist anymore. I removed this whole thing because I am able to get the OPM data out of rateData.rates in the next function.

services/ui-src/src/measures/2023/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
```diff
@@ -6,21 +6,22 @@ interface NDRforumla {
...
   ndrFormulas: NDRforumla[],
   errorLocation: string
 ) => {
   let errorArray: any[] = [];
   for (const key in rateData) {
-    if (OPM && OPM.length > 0) {
-      errorArray.push(
-        ...ComplexNoNonZeroNumOrDenom(
-          [],
-          [{ description: key, rate: [...rateData[key]["OPM"]] }],
```
angelaco11, line 19: This line was the source of the bug. `rateData[key]["OPM"]` is the legacy structure that is used in '21 and '22 measures only so that object doesn't exist and was throwing an error that blocked users from completing the measure. The new structure is `rateData["OPM"][opmLabel]` so I updated '23 and '24 to access this new structure.

BearHanded (approved):

ailZhou (approved):
Code looks good, only issue i saw was that the validation label is using the generated id instead of the label but i don't see an easy way for you to get that information. Let's just make that a separate ticket.

Screenshot 2024-07-17 at 1 50 35 PM

PR #2333, opened 2024-07-19
CMDCT-3822: removing particular measures for combined rates page and updating helper text to reflext the removed measures

angelaco11:
### Description
Simple PR to filter out measures that should not be displayed on combined rates page and then updated helper text to reflect removed measures

---
### How to test
1. On coreset page, select View Combined Rates
2. Confirm that the text under Core Set Measures Combined Rates is updated to what is in the [ticket](https://jiraent.cms.gov/browse/CMDCT-3822)
3. For Child Core Set tab, make sure that CPC-CH, LBW-CH and LRCD-CH measures are no longer displayed in the list
4. For Adult Core Set tab, make sure that CPA-AD, and NCIIDD-AD measures are no longer displayed in the list

### DIFF
--- a/services/ui-src/src/views/CombinedRatesPage/index.tsx
+++ b/services/ui-src/src/views/CombinedRatesPage/index.tsx
@@ -9,6 +9,7 @@ import { MeasureTableItem } from "components/Table/types";
 import { useFlags } from "launchdarkly-react-client-sdk";
 
 const measuresWithoutPerformanceData = [
+  "CSQ",
   "CPC-CH",
   "LBW-CH",
   "LRCD-CH",
@@ -55,10 +56,7 @@ const GetMeasuresByCoreSet = (coreSet: string, state: string, year: string) => {
     ?.filter(
       // filter out the coreset qualifiers (CSQ) and also filter out
       // the measures that are not asked to report performance measure data
-      (item) =>
-        item.measure &&
-        item.measure !== "CSQ" &&
-        !measuresWithoutPerformanceData.includes(item.measure)
+      (item) => !measuresWithoutPerformanceData.includes(item.measure)
     )
     .sort((a, b) => a?.measure?.localeCompare(b?.measure));
 
--- a/services/ui-src/src/views/CombinedRatesPage/index.tsx
+++ b/services/ui-src/src/views/CombinedRatesPage/index.tsx
@@ -45,8 +45,15 @@ const GetMeasuresByCoreSet = (coreSet: string, state: string, year: string) => {
   const measures = data?.Items as MeasureData[];
   const formatted = measures
     ?.filter(
-      // filter out the coreset qualifiers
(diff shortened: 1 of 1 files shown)


services/ui-src/src/views/CombinedRatesPage/index.tsx
```diff
@@ -45,8 +45,15 @@ const GetMeasuresByCoreSet = (coreSet: string, state: string, year: string) => {
   const measures = data?.Items as MeasureData[];
   const formatted = measures
     ?.filter(
-      // filter out the coreset qualifiers
-      (item) => item.measure && item.measure !== "CSQ"
+      // filter out the coreset qualifiers and measures where states aren't asked to report measure data
+      (item) =>
```
BearHanded, line 49: I think this could be more readable if we had a constant array stored, and this checked to make sure that the item.measure was not in that array.

services/ui-src/src/views/CombinedRatesPage/index.tsx
```diff
@@ -113,8 +120,15 @@ export const CombinedRatesPage = () => {
           Core Set Measures Combined Rates
         </CUI.Heading>
         <CUI.Text>
-          Instructions for the user - includes how to interpret the page and
-          what they need to do to see rates (i.e. complete all measures)
+          Click into a measure below to preview the preliminary combined
+          Medicaid and CHIP rate. Please complete the measure in both the
+          Medicaid and CHIP reports to ensure the combined rate is complete.
+          <br />
+          <br />
```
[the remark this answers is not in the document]
angelaco11: yeah great call, I was being super lazy. updated to using two chakra Text blocks and it formats as expected

BearHanded (approved):

services/ui-src/src/views/CombinedRatesPage/index.tsx
```diff
@@ -8,6 +8,15 @@ import { Link, useParams, useSearchParams } from "react-router-dom";
 import { MeasureTableItem } from "components/Table/types";
 import { useFlags } from "launchdarkly-react-client-sdk";
 
+const measuresWithoutPerformanceData = [
```
ailZhou, line 11: This looks good but I have one request. Can you remove NCIIDD-AD, LBW-CH & LRCD-CH as being hardcoded in the array. These measures are autocompleted and can be filtered out by the checking the autocomplete key. 
Screenshot 2024-07-23 at 10 11 57 AM

This will also cut down on us having to update the array later on if future measure gets set to be autocompleted and stops having a combine rate.

angelaco11: Just updated my PR to reflect this! Let me know if the comment and solution look good to you.

ailZhou (approved):
lgtm! thank you for making the update!

rocio-desantiago (approved):

benmartin-coforma (approved):

PR #2338, opened 2024-07-23
CMDCT-3821: removing combined rates page for states that do not have combined rates

angelaco11:
### Description
Certain states do not have separate medicaid and chip measures so they do not need access to combined rates page. This PR removes the "View Combined Rates" button from those state coreset pages.

It also gives users from said states an unauthorized error if they happen to navigate manually to the /combined-rates page:
Screenshot 2024-07-23 at 10 32 26 AM

---
### How to test
1. Log in as adminuser@test.com
2. Select any of the following states: AK, AS, DC, GU, HI, NH, NM, NC, ND, OH, PR, SC, VI, VT or WY
3. Confirm that when you get to the state coreset homepage, there is no "View Combined Rates" button
4. Navigate manually to `http://<url>/<STATE>/2024/combined-rates` and make sure that you see the unauthorized error in the screenshot above

### DIFF
--- a/services/ui-src/src/views/CombinedRatesPage/index.test.tsx
+++ b/services/ui-src/src/views/CombinedRatesPage/index.test.tsx
@@ -1,15 +1,16 @@
 import { fireEvent, render, screen } from "@testing-library/react";
 import { axe, toHaveNoViolations } from "jest-axe";
 import { useApiMock } from "utils/testUtils/useApiMock";
-import { RouterWrappedComp } from "utils/testing";
+import { BrowserRouter as Router } from "react-router-dom";
 import { CombinedRatesPage } from "views";
+import { measureDescriptions } from "measures/measureDescriptions";
 expect.extend(toHaveNoViolations);
 
 jest.mock("react-router-dom", () => ({
   ...jest.requireActual("react-router-dom"),
-  useParams: () => ({
+  useParams: jest.fn().mockReturnValue({
     year: "2024",
-    state: "OH",
+    state: "AL",
   }),
 }));
 
@@ -26,8 +27,7 @@ const useGetMeasuresValues = {
         lastAltered: 1642167976771,
         compoundKey: "AL2024ACSIET-AD",
         measure: "IET-AD",
-        description:
-          "Initiation and Engagement of Alcohol and Other Drug Abuse or Dependence Treatment",
+        description: measureDescriptions[2024]["IET-AD"],
         state: "AL",
         status: "incomplete",
         year: 2024,
@@ -39,8 +39,7 @@ const useGetMeasuresValues = {
         lastAltered: 1642167976771,
         compoundKey: "AL2024ACSAAB-AD",
         measure: "AAB-AD",
-        description:
-          "Avoidance of Antibiotic Treatment for Acute Bronchitis/Bronchiolitis: Ages 3 Months to 17 Years",
+        description: measureDescriptions[2024]["AAB-AD"],
         state: "AL",
         status: "incomplete",
         year: 2024,
@@ -54,9 +53,9 @@ describe("Test Combined Rates Page", () => {
     const apiData = { useGetMeasuresValues: useGetMeasuresValues };
     useApiMock(apiData);
     render(
-      <RouterWrappedComp>
+      <Router>
         <CombinedRatesPage />
-      </RouterWrappedComp>
+      </Router>
     );
   });
   it("renders", () => {
@@ -90,9 +89,9 @@ describe("Test accessibility", () => {
     const apiData = {};
     useApiMock(apiData);
     const { container } = render(
-      <RouterWrappedComp>
+      <Router>
         <CombinedRatesPage />
-      </RouterWrappedComp>
+      </Router>
     );
     expect(await axe(container)).toHaveNoViolations();
   });
--- a/services/ui-src/src/views/CombinedRatesPage/index.tsx
+++ b/services/ui-src/src/views/CombinedRatesPage/index.tsx
@@ -7,6 +7,7 @@ import { measureDescriptions } from "measures/measureDescriptions";
 import { Link, useParams, useSearchParams } from "react-router-dom";
 import { MeasureTableItem } from "components/Table/types";
 import { useFlags } from "launchdarkly-react-client-sdk";
+import { statesWithoutCombinedRates } from "views/StateHome";
 
 const GetColumns = () => {
   return [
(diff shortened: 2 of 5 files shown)


services/ui-src/src/views/CombinedRatesPage/index.tsx
```diff
@@ -92,6 +93,18 @@ export const CombinedRatesPage = () => {
     return <QMR.LoadingWave />;
   }
 
+  // block display from states that do not have combined rates
+  if (state && statesWithoutCombinedRates.includes(state)) {
+    return (
+      <CUI.Box data-testid="unauthorized-container">
+        <QMR.Notification
+          alertStatus="error"
+          alertTitle="You are not authorized to view this page"
```
angelaco11, line 102: Should we give them a better error than the generic unauthorized error? Something like `Combined rates for this state are not supported` or something like that?

services/ui-src/src/views/CombinedRatesPage/index.test.tsx
```diff
@@ -54,9 +53,9 @@ describe("Test Combined Rates Page", () => {
     const apiData = { useGetMeasuresValues: useGetMeasuresValues };
     useApiMock(apiData);
     render(
-      <RouterWrappedComp>
+      <Router>
```
angelaco11, line 56: I had to change from `RouterWrapperComp` to regular `Router` because `RouterWrapperComp` does a `useParams` mock return value that overrides the one in this file and makes the measure state `OH` and the year `2021`.  Since this PR blocks OH from having access to combined rates, I needed to have a useParams mock return value with a state and year that makes sense for combined rates (I chose AL and 2024).

services/ui-src/src/views/CombinedRatesPage/index.tsx
benmartin-coforma: In my opinion, yes! And since this is a page you can only navigate to by crafting an invalid URL (which is to say, this message will never be seen during user flow), the precise wording of the message shouldn't require a lot of consideration. Just enough to make sure we never get a help desk ticket for "why can't I access the combined rates page for Wissachussets"

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -39,11 +39,36 @@ interface IRepYear {
...
+  "AK",
+  "AS",
+  "DC",
+  "GU",
+  "HI",
+  "NH",
+  "NM",
+  "NC",
+  "ND",
+  "CNMI",
```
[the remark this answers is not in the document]
benmartin-coforma: It is not! I vote we omit it here.

The admin state list dropdown is populated from here: https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/master/services/ui-src/src/utils/constants.ts

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -39,11 +39,36 @@ interface IRepYear {
...
   const navigate = useNavigate();
   const { state, year } = useParams();
   const { data: reportingYears } = useGetReportingYears();
   const releasedTwentyTwentyFour = useFlags()?.["release2024"];
+  // Certain states do not have separate chip and medicaid so we will not
+  // display the Combined Rates button for those states
+  const showCombinedRatesButton =
+    state && !statesWithoutCombinedRates.includes(state);
+
+  console.log(state);
```
benmartin-coforma, line 71: Stray console log here.

```suggestion
```

services/ui-src/src/views/StateHome/index.tsx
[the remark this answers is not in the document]
benmartin-coforma: In fact, this list could very well belong in the same `constants.ts` file. Decorated with your comment below, I would say.

```ts
  /**
   * Certain states do not separate chip and medicaid,
   * so we will not display the Combined Rates button for those states.
   */
  const statesWithoutCombinedRates = [
  // ...
```

benmartin-coforma (dismissed):
Looks fine to me! I left some suggestions, which you can take or leave.

services/ui-src/src/views/StateHome/index.tsx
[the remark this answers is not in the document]
angelaco11: Oh NOICE thank you I was looking for a constants file in the ui-src root and didn't find one and clearly refused to do a search for one haha

services/ui-src/src/views/CombinedRatesPage/index.tsx
angelaco11: Wissachussets 😆

angelaco11: I agree, it would help to have just even a vaguely specific unauth error here. I updated this message to be `Combined rates for ${state} are not supported`. Let me know if anyone has any better ideas on the wording. We could get crazy about saying that "This state does not have separate medicaid and chip measures so therefore does not support combined rates" or something but I feel like that's pretty overkill

services/ui-src/src/views/StateHome/index.tsx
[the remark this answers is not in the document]
angelaco11: Moved this const to the constants file, thanks for that!

services/ui-src/src/views/StateHome/index.tsx
angelaco11: Good call, I removed this!

services/ui-src/src/views/CombinedRatesPage/index.tsx
```diff
@@ -7,7 +7,7 @@ import { measureDescriptions } from "measures/measureDescriptions";
 import { Link, useParams, useSearchParams } from "react-router-dom";
 import { MeasureTableItem } from "components/Table/types";
 import { useFlags } from "launchdarkly-react-client-sdk";
-import { statesWithoutCombinedRates } from "views/StateHome";
+import { statesWithoutCombinedRates } from "utils";
```
rocio-desantiago, line 10: would the from import be the same as 
`from "../utils/constants.ts"`

rocio-desantiago (approved):
Just a question about the import format. Not familiar with that style.

services/ui-src/src/views/CombinedRatesPage/index.tsx
BearHanded: In regard to not directly importing the contants file, `/utils` has an index file. By convention, if the referenced path is a directory, it checks for an index file, which in this case contains `export * from "./constants"`, and the import gets the reference it is looking for.

In regard to the pathing, utils is at the root of src, and you don't always need to navigate back up the directory tree, with ../, you can alternatively start at the root of the react project.

angelaco11: what Brendan said and also this was an autofill from vscode if that helps 😆

rocio-desantiago: That's cool. I've not seen this format before. It looks much cleaner!

benmartin-coforma (approved):
Looks great!

ailZhou (approved):
looks good and i can finally exercise my power of bypassing code climate!

PR #2340, opened 2024-07-23
CMDCT-3853: updating performance measure warning when checkbox is checked but data isnt filled out

angelaco11:
### Description
updating performance measure warning when checkbox is checked but num/denom/rate data for 65 and older isn't filled out

---
### How to test
1. Navigate to any 2024 measure and fill out the form. When at Definition of Population Included in the Measure section, make sure to select the `Individuals Dually Eligible for Medicare and Medicaid`
2. When filling out the numerator/denominator/rate section, DO NOT fill out the Age 65 and Older section
3. Finish filling out the form and hit the Validate button
4. You should now see the following error with the exact wording on the screenshot: 
Screenshot 2024-07-23 at 3 44 44 PM


ailZhou (approved):
lgtm! tested in deploy and the warning is only showing when the checkbox has been checked and the rate isn't filled

benmartin-coforma (approved):

rocio-desantiago (approved):
Nice work on combining these two related tickets!

PR #2345, opened 2024-07-25
ComplexNoNonZeroNumOrDenom Validation - Incorrect Label

ailZhou:
### Description

There is an issue with displaying the correct label for a complex validation, there was no path to get the description name. This is a left over issue from when QMR was using the label as an id but now that it has real ids, it needs to point to the actual label. 

Screenshot 2024-07-17 at 1 50 35 PM

It's the same changes repeated 4 times.

**Note**: I added legacy changes to all the files because I plan to merge them into one component in the future. So while its redundant and not serving a purpose now, it will make refactoring these components easier later.

### Related ticket(s)

CMDCT-3841

---
### How to test

1) Sign into QMR using [stateuserWA@test.com](mailto:stateuserDC@test.com)
2) Add a Health Home CoreSet
3) Go into the AIF-HH measure
4) In the **Measure Specification** Section, select Other.
5) In the **Data Source** Section, check Other Data Source. This will allow you to edit the rate value in the N/D/R set
6) Scroll down to **Other Performance Measure** Section, add and fill out a description and rate.
7) Scroll down to **Optional Measure Stratification** Section, and fill out an N/D/R. Change the rate to 0.0  This is how you will trigger the validation message
Screenshot 2024-07-25 at 5 01 27 PM
8) Validate the measure.
9) The error message should have the correct label name
Screenshot 2024-07-25 at 5 02 46 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/measures/2024/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
+++ b/services/ui-src/src/measures/2024/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
@@ -24,7 +24,13 @@ export const ComplexNoNonZeroNumOrDenomOMS = (
         errorArray.push(
           ...ComplexNoNonZeroNumOrDenom(
             [],
-            [{ rate: rateData[key][opmKey] }],
+            [
+              {
+                rate: isLegacyLabel()
+                  ? rateData[opmKey]["OPM"]
+                  : rateData[key][opmKey],
+              },
+            ],
             ndrFormulas,
             `${errorLocation} - ${description}`
           )
--- a/services/ui-src/src/measures/2024/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
+++ b/services/ui-src/src/measures/2024/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
@@ -1,3 +1,6 @@
+import { cleanString, isLegacyLabel } from "utils";
+import * as DC from "dataConstants";
+
 interface NDRforumla {
   numerator: number;
   denominator: number;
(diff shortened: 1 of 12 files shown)


angelaco11 (approved):
Looks good to me, nice work! Tested and label is coming through properly.

services/ui-src/src/measures/2024/shared/globalValidations/ComplexValidations/ComplexNoNonZeroNumOrDenom/index.tsx
```diff
@@ -7,21 +10,32 @@ interface NDRforumla {
...
+          ? cleanString(description)
+          : `${DC.OPM_KEY}${cleanString(description)}`;
+
         errorArray.push(
           ...ComplexNoNonZeroNumOrDenom(
             [],
-            [{ rate: rateData[key][opmLabel] }],
+            [
+              {
+                rate: isLegacyLabel()
```
angelaco11, line 29: This is more a question out of curiosity/best practices and not any kind of criticism or request for change:

Since every year has its own `ComplexNoNonZeroNumOrDenom` files and functions, is it totally necessary account for legacy data structure in years 23/24? and then also do we need to account for newer rata data in 21/22? Or do we not want to get into trying to guess which years are legacy and which ones aren't and just have it switch based on that isLegacyLabel function that consistently returns true for 21/22 and false for 23/24? I totally understand if that's the case, just wanted to hear your thoughts on it.

ailZhou: it is not necessary at all for its current state. i just did the extra work because i plan to consolidate these files in a future refactor so its easier for me to add these changes now. It could be months before i really get to the refactor and i'm likely going to forget how to structure it to handle both legacy and current code. i don't expect any other dev to go along with my side project, so you don't have to take that into account for future changes.

angelaco11: That is perfectly reasonable! I actually figured that was the case and I fully support your initiative. If I ever remember while I'm in qmr, I will attempt to take this future refactoring into account!

BearHanded (approved):
Sent a question about ways to reduce repeating our code here but this looks reasonable

PR #2350, opened 2024-07-26
Clean Up: Label to Id Migration Garbage

ailZhou:
### Description

Since we did the label to id migration for each year in parts, there are pieces of temporary code lying around can now be removed. 

What changed and why:
1. stringToLabelData function - This was use to convert string[] to LabelData[] but now that everything is a LabelData[], we do not need this function anymore
2. rateLabelText.ts (2021) - Ben had fixed it in 2022, but we don't need to create a new object to convert it to a LabelData object. The only id that needed to be fixed was "FUA-HH", as all other id matched its converted label to id value
    - I also moved the unit test to the same folder as `labels/RateLabelTexts.ts` that way the unit test can run against all the rateLabelText files instead of one.
3. I can't spell
Screenshot 2024-07-26 at 3 37 14 PM

### Related ticket(s)

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


BearHanded (approved):

services/ui-src/src/measures/2021/rateLabelText.ts
```diff
@@ -1465,20 +1465,16 @@ export const data = {
...
+export const data: RateLabelDataShape = rawData;
 
-  return {
-    qualifiers,
-    categories,
-  };
-};
\ No newline at end of file
+export const getCatQualLabels = (measure: MeasureAbbreviation) => {
+    return data[measure];
```
angelaco11, line 1479: sooo much cleaner 🔥

angelaco11 (approved):
Code logic looks great, nice cleanup. I also did some smoke testing and everything looks to be functioning it should be!

PR #2351, opened 2024-07-29
CMDCT-3854: Adding mandatory column to 2024 measures table

angelaco11:
### Description
Adding mandatory column with mandatory chip to 2024 measures
Screenshot 2024-07-29 at 1 05 17 PM

CMDCT-3854

---
### How to test
1. Navigate to 2024 coreset and select any adult measure
2. Make sure that the Mandatory column is present and that there are Mandatory chips for the following measures:
```
AMM-AD
CDF-AD
FUA-AD
FUH-AD
FUM-AD
HPCMI-AD
IET-AD
MSC-AD
OUD-AD
SAA-AD
SSD-AD
```

3. Navigate back to 2024 coreset and select child measures. Make sure that every child measure has the mandatory chip
4. Do step 3 for HH measures
5. Make sure that the mandatory column is not present for years 21, 22 and 23

### DIFF
--- a/services/app-api/types.ts
+++ b/services/app-api/types.ts
@@ -39,10 +39,12 @@ export interface Measure {
   year: number;
   placeholder?: boolean;
   /**
-   * The `autoCompleted` property is not present on measures in the database;
-   * it is set on fetch, according to the metadata in measureList.ts.
+   * The `autoCompleted` and `mandatory` properties are not present on
+   * measures in the database; they are set on fetch, according to the
+   * metadata in measureList.ts.
    */
   autoCompleted?: boolean;
+  mandatory?: boolean;
   data?: {
     DataSource?: DataSource[];
     DataSourceSelections?: unknown;
--- a/services/ui-src/src/components/Table/columns/measures.tsx
+++ b/services/ui-src/src/components/Table/columns/measures.tsx
@@ -83,6 +83,25 @@ export const measuresColumns = (
         );
       },
     },
+    // {
+    //   header: "Mandatory",
+    //   id: "mandatory_column_header",
+    //   styleProps: { textAlign: "center" },
+    //   cell: (data: MeasureTableItem.Data) => {
+    //     console.log(data);
+    //     return (
+    //       <CUI.Badge
+    //         fontSize="xs"
+    //         colorScheme="blue"
+    //         textTransform="capitalize"
+    //         borderRadius="lg"
+    //         px="2"
+    //       >
+    //         <CUI.Text fontWeight="normal">Mandatory</CUI.Text>
+    //       </CUI.Badge>
+    //     );
+    //   },
+    // },
     {
       header: "Reporting FFY " + year,
       id: "reporting_column_header",
--- a/services/ui-src/src/components/Table/columns/measures.tsx
+++ b/services/ui-src/src/components/Table/columns/measures.tsx
@@ -55,6 +55,7 @@ const MeasureStatusText = ({
 export const measuresColumns = (
   year: string
 ): TableColumn<MeasureTableItem.Data>[] => {
+  const displayMandatoryColumn = year === "2024";
   return [
     {
       header: "Abbreviation",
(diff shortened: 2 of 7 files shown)


services/ui-src/src/components/Table/columns/measures.tsx
```diff
@@ -83,6 +84,30 @@ export const measuresColumns = (
         );
       },
     },
+    ...(displayMandatoryColumn
```
angelaco11, line 87: This is a really weird implementation to conditionally render an element in an array, but in my opinion it was the cleanest way to conditionally render this column for 2024 measures. I am fully open to any suggestions if anyone has a better idea!!!

services/ui-src/src/components/Table/columns/measures.tsx
```diff
@@ -55,6 +55,7 @@ const MeasureStatusText = ({
 export const measuresColumns = (
   year: string
 ): TableColumn<MeasureTableItem.Data>[] => {
+  const displayMandatoryColumn = year === "2024";
```
angelaco11, line 58: Is this an acceptable way to render stuff for 2024? I saw this in other parts of the application such as with rendering the `View Combined Rates` button and I couldn't think of a better way to do it.

BearHanded: this at least needs to be revisited when we hit 2025. I would advise at minimum year >= 2024, and with that said think we need to start tracking some features by year in a centralized place. I think we have a lot of these across the app.

Maybe we centralize a featureByYear file that contains all of the yearly behaviors, and for each function/flag returns true or false based off the provided year?

angelaco11: Yeah i really like the idea of a `featureByYear` file almost like how we use the commonLabels files for each year in the questionnaires. 

Should we make another ticket for that? Anyone who picks that up could go through and find all the places where we are doing stuff like `year === 2024` and replace it with new `featureByYear` stuff. Would be a bit of an undertaking but worth it in my opinion, especially when we get into 2025 and beyond

ailZhou: the only other way i can see doing this without using the year is to check whether mandatory even exist in the measure and then filtering that column generation out if it doesn't. 

```
  let columns = QMR.measuresColumns(year);
  if(!measures.some((measure) => measure.mandatory))
    columns = columns.filter((column) => column.header != "Mandatory");
```
But this only works if there's at least one measure that is mandatory in a mandatory coreset 😓

ailZhou: @angelaco11 just saw your response to Brendan. That ticket would probably end up as being part of the refactoring, so if you make it, just toss it into the QMR Refactor Board.

angelaco11: Ok yeah I feel like creating that ticket and chuckin on the refactor board sounds most reasonable here. You guys cool with that conclusion?

angelaco11: here is said ticket for vis https://jiraent.cms.gov/browse/CMDCT-3883

services/app-api/types.ts
```diff
@@ -39,10 +39,12 @@ export interface Measure {
   year: number;
   placeholder?: boolean;
   /**
-   * The `autoCompleted` property is not present on measures in the database;
-   * it is set on fetch, according to the metadata in measureList.ts.
+   * The `autoCompleted` and `mandatory` properties are not present on
+   * measures in the database; they are set on fetch, according to the
+   * metadata in measureList.ts.
```
benmartin-coforma, line 44: You added relevant info to my comment, and kept the grammar correct 😭 ❤️ 🙏

services/ui-src/src/components/Table/columns/measures.tsx
benmartin-coforma: I think this is a good choice. It does feel a bit awkward, but the other ideas that popped into my head would be more awkward still 
* Adding it unconditionally but filtering afterward
* Splicing it in afterward, conditionally
* Working in some kind of `shouldDisplay` flag and referencing it later? Somehow?

The advantage of the current approach is that it's 100% inline. Want to know what's in the array? Follow this process:
1. Look where the array is defined
2. You're done.

I appreciate processes like that.

benmartin-coforma (approved):

services/ui-src/src/components/Table/columns/measures.tsx
angelaco11: Thank you for giving me more ideas on how to do that! The adding it unconditionally but filtering afterward is also a super solid idea that I hadn't considered!! But I agree. with you, in this particular case where we are rendering columns, it is really nice to have it inline even though it's not the most common syntax

services/app-api/types.ts
angelaco11: your comment was just so well written that it was that easy to add on to 🫶

rocio-desantiago (approved):
Very straightforward. Great brainstorming sesh on prospective future years.

PR #2343, opened 2024-07-24
Combine Rates - Hybrid Data Source Only 

ailZhou:
### Description

This PR contains the hybrid measure calculation used in the Combined Rates section. Despite the careful planning withe administrative calculation, we still needed to do a couple of rewrites to accommodate `HybridMeasurePopulationIncluded` field we are pulling from the measures.. Its located in the **Definition of Population Included in the Measure** section.
 
Screenshot 2024-07-25 at 12 02 40 PM

**New Feature**
- The Combined Rates N/D/R table will add extra rows based on whether the data source has a hybrid selected.

**Side Issue Fix**: There was an issue with CHIP data source displaying in the data source information banner. It was caused by the naming update from CHIP to Separate CHIP

### Related ticket(s)

CMDCT-3652

---
### How to test

1) Sign into QMR, [stateuser1@test.com](mailto:stateuser1@test.com)
2) Select a Hybrid Measure to fill out, i.e. IMA-CH

Here are the possible state of the table based on what values were entered in.

State 1: No data source selected, no Measure-Eligible Population filled,  Rates filled
Screenshot 2024-07-25 at 12 06 15 PM

State 2: Medicaid data source selected, no Measure-Eligible Population filled, Rates filled 
Screenshot 2024-07-25 at 12 08 08 PM

State 3: Medicaid data source selected, Medicaid Measure-Eligible Population filled, Rates filled 
Screenshot 2024-07-25 at 12 08 56 PM

State 4: Both data source selected, Medicaid Measure-Eligible Population filled, Rates filled
Screenshot 2024-07-25 at 12 09 20 PM

State 5: Both data source selected, both Measure-Eligible Population filled, Rates filled
Screenshot 2024-07-25 at 12 09 44 PM

### Notes

N/A

---
### Pre-review checklist

- [x] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [ ] I have updated relevant documentation, if necessary
- [x] I have performed a self-review of my code
- [x] I have manually tested this PR in the deployed cloud environment

---
#### Review
- [ ] Design: This work has been reviewed and approved by design, if necessary
- [x] Product: This work has been reviewed and approved by product owner, if necessary

#### Security
_If either of the following are true, notify the team's ISSO (Information System Security Officer)._

- [ ] These changes are significant enough to require an update to the SIA.
- [ ] These changes are significant enough to require a penetration test.
---

[...]

PR #2353, opened 2024-07-30
Hybrid & Another Data Source Calculation 

ailZhou:
### Description

The Hybrid Other Cal is the same as the [Hybrid Cal](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2343).

The functionality is the same so the Hybrid Other class is extended from the Hybrid class. While we could just expand the Hybrid class data source map to include the Hybrid Other data source map, I keep it separate as the BO's seem to see it as two separate types and we may have to expand them differently in the future.

Here is the data source map for the Hybrid Other
Screenshot 2024-07-30 at 11 34 31 AM

The biggest change is to the check function in the parent class, it wasn't doing a full match before triggering. i.e. if you had two out of three data source that matched a particular medicaid, it would consider it a match and return true. This PR expands that to check to make sure that the data source arrays have the same number of items (prevent passing if only two out of three is a match) and it sorts the array before comparing it by index (making sure that the data sources are a match if they are ordered in the same index in the array)

### Related ticket(s)

CMDCT-3653

---
### How to test

Deploy Link: https://dq4gnz40iqajr.cloudfront.net/

Same testing as [Hybrid Cal](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2343) but make sure to use the data sources listed in the table above.

1) Sign into QMR, [stateuser1@test.com](mailto:stateuser1@test.com)
2) Select a Hybrid Measure to fill out, i.e. IMA-CH

Here are the possible state of the table based on what values were entered in.

State 1: No data source selected, no Measure-Eligible Population filled,  Rates filled
Screenshot 2024-07-25 at 12 06 15 PM

State 2: Medicaid data source selected, no Measure-Eligible Population filled, Rates filled 
Screenshot 2024-07-25 at 12 08 08 PM

State 3: Medicaid data source selected, Medicaid Measure-Eligible Population filled, Rates filled 
Screenshot 2024-07-25 at 12 08 56 PM

State 4: Both data source selected, Medicaid Measure-Eligible Population filled, Rates filled
Screenshot 2024-07-25 at 12 09 20 PM

State 5: Both data source selected, both Measure-Eligible Population filled, Rates filled
Screenshot 2024-07-25 at 12 09 44 PM

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


services/app-api/handlers/rate/calculations/hybridCalculation.ts
```diff
@@ -15,7 +15,7 @@ export class HybridCalculation extends RateCalculation {
     },
   ];
 
-  private totalMeasureEligiablePopulation: number = 0;
+  protected totalMeasureEligiablePopulation: number = 0;
```
rocio-desantiago, line 18: small nit, but not blocking
`totalMeasureEligiablePopulation` to `totalMeasureEligiblePopulation` but could be done in the future or never really. 😆

benmartin-coforma (approved):
The code all makes sense 👍

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
```diff
@@ -110,10 +110,11 @@ const getRateComponent = (json: CombinedRatePayload) => {
     .map((item) => (item as SeparatedData)?.dataSource)
     .flat();
 
-  const isHybrid = dataSources?.includes(
-    "HybridAdministrativeandMedicalRecordsData"
-  );
-
+  const hybridDataSources = [
+    "HybridAdministrativeandMedicalRecordsData",
+    "Casemanagementrecordreview",
```
rocio-desantiago, line 115: Would it be a big change to update `Casemanagementrecordreview` to`CaseManagementRecordReview` to follow the PascalCase naming convention similar to `HybridAdministrativeandMedicalRecordsData`
I know it's a dataSource, so not sure if it's a big deal here.

rocio-desantiago (approved):
Non blocking comments

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
ailZhou: Unfortunately, I don't think we can change that at this point of the development cycle. The naming of Case Management Record Review was set before we inherited the project and is being ingest downstream so it might be risky to change it at this point

services/app-api/handlers/rate/calculations/hybridCalculation.ts
ailZhou: good catch, i was change it in the PCR cal branch!

PR #2355, opened 2024-07-30
Core Set Order - Home page

ailZhou:
### Description

The BOs want the coreSet in the table to be displayed in a certain order:

For Separate Reports Order should be: 
- Child Medicaid (Title XIX & XXI)
- Child Separate CHIP
- Adult Medicaid (Title XIX & XXI)
- Adult Separate CHIP

For Combined Reports Order should be: 
- Child Combined 
- Adult Combined 

As a means to not have to create a string array to order the coreSets, I have chosen to use the CoreSetAbbr enums to be the list the table items will follow. If anyone has a better idea, please let me know!

David also agreed to let me apply it to the previous years as well so we don't have to do a conditional for 2024 onward.

### Related ticket(s)

CMDCT-3880

---
### How to test

1) Sign into QMR for all these users [stateuser1@test.com](mailto:stateuser1@test.com), [stateuser3@test.com](mailto:stateuser3@test.com) & [stateuserWA@test.com](mailto:stateuserDC@test.com)

For stateuser1@test.com
Screenshot 2024-07-30 at 1 20 48 PM

For stateuser3@test.com
Screenshot 2024-07-30 at 1 21 06 PM

For stateuserWA@test.com, add the Health Home CoreSet
Screenshot 2024-07-30 at 1 21 21 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/types.ts
+++ b/services/app-api/types.ts
@@ -75,12 +75,12 @@ export interface Banner {
 }
 
 export const enum CoreSetAbbr {
-  ACS = "ACS", // adult
-  ACSM = "ACSM", // adult medicaid
-  ACSC = "ACSC", // adult chip
   CCS = "CCS", // child
   CCSM = "CCSM", // child medicaid
   CCSC = "CCSC", // child chip
+  ACS = "ACS", // adult
+  ACSM = "ACSM", // adult medicaid
+  ACSC = "ACSC", // adult chip
   HHCS = "HHCS", // health homes
 }
 
--- a/services/app-api/types.ts
+++ b/services/app-api/types.ts
@@ -75,12 +75,12 @@ export interface Banner {
 }
 
 export const enum CoreSetAbbr {
-  CCS = "CCS", // child
-  CCSM = "CCSM", // child medicaid
-  CCSC = "CCSC", // child chip
   ACS = "ACS", // adult
   ACSM = "ACSM", // adult medicaid
   ACSC = "ACSC", // adult chip
+  CCS = "CCS", // child
+  CCSM = "CCSM", // child medicaid
+  CCSC = "CCSC", // child chip
   HHCS = "HHCS", // health homes
 }
 
--- a/services/ui-src/src/types.ts
+++ b/services/ui-src/src/types.ts
@@ -1,10 +1,11 @@
+//this is also used to determine the order of the coreSets in the homepage table so ordering matters
 export enum CoreSetAbbr {
-  ACS = "ACS",
-  ACSM = "ACSM",
-  ACSC = "ACSC",
   CCS = "CCS",
   CCSM = "CCSM",
   CCSC = "CCSC",
+  ACS = "ACS",
+  ACSM = "ACSM",
+  ACSC = "ACSC",
   HHCS = "HHCS",
 }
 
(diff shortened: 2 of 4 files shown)


services/ui-src/src/types.ts
benmartin-coforma, line 1: It is very rare for the ordering of keys within an object, or of elements within an enum, to be semantically relevant. I believe that this works today, but it feels very fragile - updates to Typescript might conceivably break it, and future developers ignoring your (helpful!) comment certainly might.

I'd much rather see an explicit array declaring the order - either right here, or in the homepage. And honestly, the homepage makes more sense - since that's the place where the ordering matters.

benmartin-coforma: This might look something like this:
```ts
const CoreSetDisplayOrder = [
  CoreSetAbbr.CCS,
  CoreSetAbbr.CCSM,
  CoreSetAbbr.CCSC,
  CoreSetAbbr.ACS,
  CoreSetAbbr.ACSM,
  CoreSetAbbr.ACSC,
  CoreSetAbbr.HHCS,
];
```

services/ui-src/src/views/StateHome/index.tsx
```diff
@@ -287,6 +288,21 @@ const StateHome = () => {
...
+        ...formattedTableItems.filter((item) => item.coreSet.includes(abbr))
+      );
+      return;
+    }
+    const coreSetItem = formattedTableItems.find(
+      (item) => item.coreSet === abbr
+    );
+    if (coreSetItem) sortedTableItems.push(coreSetItem);
+  });
+
```
benmartin-coforma, line 305: Once again you've written a helpful comment - but if we were referencing `CoreSetDisplayOrder` instead of `Object.values(CoreSetAbbr)`, it would be unnecessary.

As always, I think `.forEach` is a less-than-ideal choice; `for..of` is superior in almost every way when creating a procedural loop. But in this case, I don't think we even want a procedural loop. Building up an array by pushing in elements one at a time, while referencing another array is the _exact_ time to use `.map`. The HHCS core sets add a small wrinkle we can iron out with `.flatMap` instead.

```suggestion
  const sortedTableItems = CoreSetDisplayOrder.flatMap((abbr) =>
   abbr === CoreSetAbbr.HHCS
      ? formattedTableItems.filter((item) => item.coreSet.startsWith(abbr))
      : formattedTableItems.filter((item) => item.coreSet === abbr)
  );

```

services/app-api/types.ts
benmartin-coforma, line 1: If you do decide to create an explicit `CoreSetDisplayOrder` array, this change won't be needed.

benmartin-coforma (commented):
Thank you in advance for your patience with me here, as I request something I'm sure you already considered and discarded as being too verbose 😅

services/ui-src/src/types.ts
ailZhou: i will concede as even I knew it was a stretch to use enums 😅. I'm willing to accept this solution since it doesn't involve writing the string name over again

services/ui-src/src/views/StateHome/index.tsx
ailZhou: i did something similar earlier on using map but it was less eloquent. On deploy it kept saying I had typescript issues, so I'm hoping that it was cause by me using enums and that its fine now that i've switched over to your code.

services/app-api/types.ts
ailZhou: tbf, this change was never needed. i just moved it to have it match

BearHanded (approved):

benmartin-coforma (approved):

PR #2360, opened 2024-07-31
Measure Specific Calculation - PCR-AD

ailZhou:
### Related ticket(s)

CMDCT-3878

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/app-api/types.ts
+++ b/services/app-api/types.ts
@@ -25,6 +25,12 @@ export interface StandardRateShape {
   ["weighted rate"]?: string;
 }
 
+export interface StandardValueShape {
+  label: string;
+  uid?: string;
+  value?: string;
+}
+
 export interface Measure {
   compoundKey: string;
   coreSet: CoreSetAbbr;
--- a/services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
+++ b/services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
@@ -20,6 +20,7 @@ type TableKeys = {
   uid: string;
   label: string;
   category?: string;
+  type: string;
 };
 /** An intermediate shape, not guaranteed to have an entry for every program type */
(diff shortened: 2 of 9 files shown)


services/app-api/types.ts
```diff
@@ -18,13 +18,22 @@ export interface CoreSet {
 export interface StandardRateShape {
   label: string;
   uid?: string;
```
BearHanded, line 20: is "type" also a prop here now, or am I misinterpreting the below logic

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
```diff
@@ -230,6 +302,7 @@ function provideDefaultValues(
         rate,
         ["measure-eligible population"]: mep,
         ["weighted rate"]: weightRate,
```
BearHanded, line 304: these keys still make me a bit uncomfortable

services/app-api/types.ts
ailZhou: nope, type is determine based on property call in the front end so there's no api layer prop being passed to the front

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
ailZhou: i do plan to refactor this file when i come back now that it has to expand to handle a different field type than we expected so not using string as key names can be a part of that

services/app-api/types.ts
BearHanded: should there be and would that make things simpler? It at least would give anyone looking at the data downstream a hint as to what they're parsing

BearHanded (approved):
We will make a pass refactoring this but this behaves as promised and gets us to testing next week.

ajaitasaini (approved):

PR #2373, opened 2024-08-16
CMDCT-3929: making definition of denom text changes for not separated coresets

angelaco11:
### Description
Super simple PR to update options for coresets that don't have separate medicaid and chip measures

### Related ticket(s)
CMDCT-3929

---
### How to test
1. Log in as a user with state that doesn't have separate coresets for medicaid and chip (such as stateuserDC)
2. Navigate to adult coreset and click on any measure and scroll to Definition of denominator section and make sure it matches the updates noted in the [ticket](https://jiraent.cms.gov/browse/CMDCT-3929)
3. repeat step 2 with the child coreset

### DIFF
--- a/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
+++ b/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
@@ -170,6 +170,10 @@ const coreSetSpecificOptions: CoreSetSpecificOptions = {
         value: DC.DENOMINATOR_INC_MEDICAID_POP,
       },
       {
+        displayValue: "Medicaid-Expansion CHIP (Title XXI)",
+        value: DC.DENOMINATOR_INC_MEDICAID_EXPANSION,
+      },
+      {
         displayValue: "Individuals Dually Eligible for Medicare and Medicaid",
         value: DC.DENOMINATOR_INC_MEDICAID_DUAL_ELIGIBLE,
       },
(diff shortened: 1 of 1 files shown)


BearHanded (approved):
verified, looks good

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -178,11 +182,13 @@ const coreSetSpecificOptions: CoreSetSpecificOptions = {
       <>
         <CUI.Text mt="3">
           Please select all populations that are included in the denominator.
-          For example, if your data include both Medicaid (Title XIX) enrollees
-          and individuals dually eligible for Medicare and Medicaid, select:
+          For example, if your data include Medicaid (Title XIX) beneficiaries,
```
rocio-desantiago, line 185: If my engrish is right, I think this should say: 
"if your data includes Medicaid..."

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -197,22 +203,20 @@ const coreSetSpecificOptions: CoreSetSpecificOptions = {
...
+        value: DC.DENOMINATOR_INC_MEDICAID_EXPANSION,
       },
     ],
     helpText: (
       <>
         <CUI.Text mb="2">
           Please select all populations that are included in the denominator.
-          For example, if your data include both Medicaid (Title XIX) enrollees
-          and individuals dually eligible for Medicare and Medicaid, select:
+          For example, if your data include both Medicaid (Title XIX) and
```
rocio-desantiago, line 214: The grammar is hard for me on this one since data can be both singular or plural. It's just scratching my brain a bit reading it as is. 😆

rocio-desantiago (approved):
`data` is plural, so "include" is appropriate 😄

PR #2374, opened 2024-08-16
CMDCT-3927: removing MSC-AD measure from combined rates page

angelaco11:
### Description
The simplest PR to remove MSC-AD from combined rates list

### Related ticket(s)
CMDCT-3927

---
### How to test
1. Log in as user with a state that has combined rates
2. Navigate to combined rates page, click on the Adult tab and make sure that MCS-AD measure is not in the list

### DIFF
--- a/services/ui-src/src/measures/2024/CPAAD/index.tsx
+++ b/services/ui-src/src/measures/2024/CPAAD/index.tsx
@@ -33,7 +33,7 @@ export const CPAAD = ({
         <>
           <Q.HowDidYouReport />
           <Q.DataSource type="adult" />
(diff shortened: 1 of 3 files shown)


BearHanded (approved):
Validated in deployed env

BearHanded (approved):

rocio-desantiago (approved):
Confirmed it's not on the Adult Core Set of the Combined Rates Page and the language was updated at the top of the page.

### Language updated
Combined_Rates_Page_Language-updated

### MSC-AD removed from list
MSC_AD-removed-from-list

PR #2377, opened 2024-08-16
CMDCT-3928: switching CPA-AD measure to use its own custom DefinitionOfPopulation

angelaco11:
### Description
This is a tragic PR. Because of the changes requested in CMDCT-3928, we need to go back to using a custom DefinitionOfPopulation specifically for CPA-AD measure. Previously, Aileen had successfully switched all the measures over to using the common DefinitionOfPopulation component so this sets us back a bit on that front. 

@BearHanded idk what you think about this. we could TECHNICALLY add the CPA-AD specific definitions to the common file, but that would make an already bloated component even more bloated and lead to some potentially painful conditional rendering.

### Related ticket(s)
CMDCT-3928

---
### How to test
1. Log in a user with separate medicaid and chip measures
2. Go to adult medicaid coreset and click on the CPA-AD measure
3. Scroll to Definition of Denominator section and make sure that the wording is what matches what is requested in the [ticket](https://jiraent.cms.gov/browse/CMDCT-3928)
4. Repeat steps 2 & 3 for chip coreset

### DIFF
--- a/services/ui-src/src/measures/2024/CPAAD/index.tsx
+++ b/services/ui-src/src/measures/2024/CPAAD/index.tsx
@@ -33,7 +33,7 @@ export const CPAAD = ({
         <>
           <Q.HowDidYouReport />
           <Q.DataSource type="adult" />
-          <CMQ.DefinitionOfPopulation />
+          <Q.DefinitionOfPopulation coreSetId={coreSetId as string} />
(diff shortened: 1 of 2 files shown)


rocio-desantiago (commented):
Unit test is failing after 
src/measures/2024/IUHH/index.test.tsx

PR #2380, opened 2024-08-19
CMDCT-3928: Adding specific language to definition of denom for CPA-AD measure because why not

angelaco11:
### Description
CPA-AD needs custom definition of denominator options and helper text. I did this by adding custom options to the CPA-AD index file and passing them in as an optional prop to DefinitionsOfPopulation common component. Then either render the prop or default coreset specific ones. Thanks to @ailZhou for that solution!

### Related ticket(s)
CMDCT-3928

---
### How to test
1. Login as user with separate medicaid and chip coresets
2. Navigate to adult medicaid and to the CPA-AD measure
3. Make sure that the text under "Definition of Denominator" matches what is in the [ticket](https://jiraent.cms.gov/browse/CMDCT-3928)
4. Repeat steps 2 & 3 for adult chip

### DIFF
--- a/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
+++ b/services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
@@ -11,79 +11,25 @@ import * as DC from "dataConstants";
 import { useContext } from "react";
 import SharedContext from "shared/SharedContext";
 import { AnyObject } from "types";
-import { useParams, useLocation } from "react-router-dom";
+import { useParams } from "react-router-dom";
 
 interface Props {
   childMeasure?: boolean;
   hybridMeasure?: boolean;
   populationSampleSize?: boolean;
   healthHomeMeasure?: boolean;
+  coreSetOptions?: CoreSetSpecificOptions;
 }
 interface DefOfDenomOption {
   displayValue: string;
   value: string;
 }
-interface CoreSetSpecificOptions {
+export interface CoreSetSpecificOptions {
   [coreSetId: string]: {
     options: DefOfDenomOption[];
     helpText: JSX.Element;
   };
 }
-
-const surveyMeasures = ["CPA-AD"];
-
-const surveyMeasureOptions: CoreSetSpecificOptions = {
-  ACSM: {
-    options: [
-      {
-        displayValue: "Medicaid (Title XIX)",
-        value: "Medicaid (Title XIX)",
-      },
-      {
-        displayValue: "Medicaid-Expansion CHIP (Title XXI)",
-        value: "SurveySampleIncCHIP",
-      },
-      {
-        displayValue: "Individuals Dually Eligible for Medicare and Medicaid",
-        value: "SurveySampleIncMedicareMedicaidDualEligible",
-      },
-    ],
-    helpText: (
-      <>
-        <CUI.Text mt="3" mb="3">
-          Please select all populations that are included in the survey sample.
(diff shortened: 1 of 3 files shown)


ajaitasaini:
it seems like DefinitionOfPopulation is called / used in each individual measure.. sort of cumbersome but you could pass in the measureName as a prop? 
Screenshot 2024-08-19 at 5 05 19 PM

(this is PCRAD/index.tsx)

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -405,8 +465,14 @@ export const DefinitionOfPopulation = ({
   const labels: any = useContext(SharedContext);
 
   const params = useParams();
+  const { pathname } = useLocation();
   const coreSetType = healthHomeMeasure ? "HHCS" : params.coreSetId;
 
+  const measureName = pathname.substring(
```
rocio-desantiago, line 471: is it possible to use the existing measureName property that Ajaita mentioned? Not sure how doable that is or if it can be applied for this case.

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -405,8 +465,14 @@ export const DefinitionOfPopulation = ({
...
+    pathname.lastIndexOf("/") + 1,
```
rocio-desantiago, line 472: this might be more readable like this:
`const measureName = pathname.split("/").pop()

rocio-desantiago (commented):
measureName logic

ailZhou (approved):
So i thought very hard on an alternative and besides the suggestions from @ajaitasaini & @rocio-desantiago, the other options would require a refactor of definitionsOfPopulation into a class to be extended by cpa-ad which is too much work for this and might end up getting changed in a refactor anyway.

The only other smaller options that I can think of is instead of passing the measureName a prop, you pass the `surveyMeasureOptions` as a prop and use that to override `coreSetSpecificOptions`. So you would store the options in like `.../2024/CPAAD/index.tsx` and just pass it as prop there. It cuts out needing to know the measureName and allow other measures to override those values easier but more props.

i'm still going to approve your PR because what you wrote is still good as it is and i have no qualms about using paths.

rocio-desantiago (dismissed):
Works as-is. I'll keep an eye out if there are any changes to the code based on feedback

angelaco11:
> So i thought very hard on an alternative and besides the suggestions from @ajaitasaini & @rocio-desantiago, the other options would require a refactor of definitionsOfPopulation into a class to be extended by cpa-ad which is too much work for this and might end up getting changed in a refactor anyway.
> 
> The only other smaller options that I can think of is instead of passing the measureName a prop, you pass the `surveyMeasureOptions` as a prop and use that to override `coreSetSpecificOptions`. So you would store the options in like `.../2024/CPAAD/index.tsx` and just pass it as prop there. It cuts out needing to know the measureName and allow other measures to override those values easier but more props.
> 
> i'm still going to approve your PR because what you wrote is still good as it is and i have no qualms about using paths.

Yeah I really like this idea. I would prefer to keep the common component measure agnostic and include the options specific to cpa-ad in the cpa-ad measure component. Then the common component just overrides the options if they exist from the parent component. I'm currently cookin something up based on your suggestion

ailZhou (approved):

BearHanded (approved):
Assuming this gets the thumbs up from David, it's good to go

PR #2381, opened 2024-08-19
CMDCT-3918: adding gross ternary to render health home question

angelaco11:
### Description
This is gross and I hate it. Basically adding a healthHomeQuestion label to 2024 labels and rendering that if it exists and is a health home measure. otherwise rendering the question

### Related ticket(s)
CMDCT-3918

---
### How to test
1. Navigate to user with HH measures (DC is one)
2. Click into any health home measure and make sure that the "Does this denominator represent..." question has the same wording that is requested in the [ticket](https://jiraent.cms.gov/browse/CMDCT-3918)
3. Click into any adult measure and make sure the "Does this denominator represent..." question does not reference health home

### DIFF
--- a/services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
+++ b/services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
@@ -68,6 +68,8 @@ export const commonQuestionsLabel = {
     measureEligiblePopDenom: {
       question:
         "Does this denominator represent your total measure-eligible population as defined by the technical specifications for this measure? This includes beneficiaries who move between programs (Medicaid and CHIP), plans, or delivery systems during the measurement year but met continuous enrollment requirements at the state level.",
+      healthHomeQuestion:
+        "Does this denominator represent your total measure-eligible population as defined by the technical specifications for this measure? This includes enrollees who move between health home providers, plans, or delivery systems during the measurement year but met continuous enrollment requirements at the state level.",
       optionYes:
         "Yes, this denominator includes the total measure-eligible population as defined by the Technical Specifications for this measure.",
       optionNo:
--- a/services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
+++ b/services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
@@ -66,10 +66,11 @@ export const commonQuestionsLabel = {
   DefinitionsOfPopulation: {
     defineDenomOther: "Define the other denominator population:",
     measureEligiblePopDenom: {
-      question:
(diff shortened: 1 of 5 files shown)


services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
```diff
@@ -68,6 +68,8 @@ export const commonQuestionsLabel = {
     measureEligiblePopDenom: {
       question:
```
ailZhou, line 69: Possible alternative to using the ternary? 

If we expand the question key into an object with coreSetType keys, we can avoid using the ternary.

```
question: {
        default:
          "Does this denominator represent your total measure-eligible population as defined by the technical specifications for this measure? This includes beneficiaries who move between programs (Medicaid and CHIP), plans, or delivery systems during the measurement year but met continuous enrollment requirements at the state level.",
        HHCS: "Does this denominator represent your total measure-eligible population as defined by the technical specifications for this measure? This includes enrollees who move between health home providers, plans, or delivery systems during the measurement year but met continuous enrollment requirements at the state level.",
},
```

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -445,19 +447,18 @@ export const DefinitionOfPopulation = ({
         <QMR.RadioButton
           formLabelProps={{ fontWeight: "600" }}
           label={
-            labels.DefinitionsOfPopulation.measureEligiblePopDenom.question
+            healthHomeMeasure && mepDenomLabels.healthHomeQuestion
```
ailZhou, line 450: if we do the above, we can then call it like this instead.

```
  label={
    mepDenomLabels.question[coreSetType!] ??
    mepDenomLabels.question.default
  }
```

ailZhou (dismissed):
left some comments on a possible alternative to using the ternary. what you wrote is good too so i'm approving if you don't want to change it that way!

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
angelaco11: dayum you are a genius. Updated to use this much better logic, thank you!!

services/ui-src/src/labels/2024/commonQuestionsLabel.tsx
angelaco11: this is really great thinking, I made these updates!!

ailZhou (approved):
lgtm! tested the changes on deploy, thank you for making the update 🎉

rocio-desantiago (approved):
Tested in deploy link and the content is correct for both Health Home and Adult - Non Home Health measures.

PR #2385, opened 2024-08-20
Timeout and save functionality, Again

ailZhou:
### Description

Hopefully second times the charm. 

There's a couple more false positive being triggered when it comes to whether the data has changed during a timeout. The biggest one has to do with performance measure. When we render the performance measure section, it is actually rewriting the data into the field each time we interact with it, including when we load the page (i'm mostly positive that's what is happen). This is causing react hook form to think that the data has changed so therefore will mark it as dirty. 

To combat this I have written a recursive loop to check only the data between the default value object and the data object. If there is a difference between the two, we will allow the system to save during time out, if not, we ignore it.

### Related ticket(s)

CMDCT-3634

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -244,16 +244,42 @@ export const MeasureWrapper = ({
     validateAndSetErrors(data);
   };
 
+  //compare two objects and return true if the values in them are different
+  const areObjectsDifferent = (objectA: AnyObject, objectB: AnyObject) => {
+    for (const key of Object.keys(objectA ?? objectB)) {
+      //if the value is an object or array, we want to go down another layer
+      if (typeof objectA?.[key] === "object") {
+        //once we see a difference we want to break out of the loop and return true to trigger a save
+        if (areObjectsDifferent(objectA?.[key], objectB?.[key])) {
+          return true;
+        }
+      } else {
+        if (objectA?.[key] !== objectB?.[key]) return true;
+      }
+    }
+    return false;
+  };
+
+  const hasDataChanged = (data: any) => {
+    //there are some instances where there is not change to the data but the way we load the data triggers react hook form to think there is.
+    //this function is to do a comparison between the defaultValues (prev saved data) and data (data waiting to be saved)
+    if (Object.keys(methods.formState.dirtyFields).length === 0) return false;
+
+    //instead of looping through all the data, we will loop through only the keys that react hook form indicated has a change
+    const fieldKeys = Object.keys(methods.formState.dirtyFields);
+
+    //check if any field has changed
+    return fieldKeys.some((key) =>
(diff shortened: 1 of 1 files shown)


services/ui-src/src/utils/form.ts
```diff
@@ -48,3 +49,19 @@ export const areSomeRatesCompleted = (data: any, measureId: string = "") => {
 
   return ratesExist;
 };
+
+//compare two objects and return true if the values in them are different
+export const areObjectsDifferent = (objectA: AnyObject, objectB: AnyObject) => {
+  for (const key of Object.keys(objectA ?? objectB)) {
```
BearHanded, line 55: this assumes all objects _have_ the same keys already. New props on B won't be cycled through to check against A.

ex: these two would flag as equal
```
const a = {"red": 1};
const b = {"blue": 1, "red": 1};
```

I think this needs to check the length of the keys arrays are equal before cycling

benmartin-coforma, line 55:
```suggestion
  const uniqueKeys = [...Object.keys(objectA ?? {}), ...Object.keys(objectB ?? {})]
    .filter((key, index, arr) => index === arr.indexOf(key));
  for (const key of uniqueKeys) {
```

benmartin-coforma: ^ This would do it, if the change is necessary.

I say "if" because _a_ change is certainly necessary; the function doesn't do what it says it does. But I'm not sure if we _need_ what the function says it does? If `objectA` is only the _dirty_ fields from React's form management, then maybe we don't need full equality of the objects? And if so, then the only change required is a function rename.

Or maybe we do need to change the behavior; I'm a little bit lost here to be honest.

BearHanded (approved):

services/ui-src/src/utils/form.ts
ailZhou: took Ben's suggestion and made the changes

ailZhou: so i tried to fix this once before [here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2243). originally, i think there was a bug in react hook form because `methods.formState.isDirty` would be true and `methods.formState.dirtyFields` would be empty. i swapped to using `methods.formState.dirtyFields` and just checking to see if any values have changed there.

the issue with `methods.formState.dirtyFields` is that it will say which key has changed but not tell me what has changed, it would just return a boolean like so:
```
{
    "PerformanceMeasure": {
        "rates": {
            "ZCy3XP": true
        }
    }
}
```
not exactly very helpful so now i want to compare the two values of what has been saved and what will be saved instead because i don't trust react hook form (although this time it is my fault, its because the N/D/R sets without categories were trying to save an undefined label). 

If this doesn't work and the client comes back for a third time, i'm going to insist very intensely that we take out autosave.

benmartin-coforma (approved):

PR #2386, opened 2024-08-20
Refactor Combined Rates

benmartin-coforma:
### Description
There are a ton of changes in here, but the driving force is these three big ones:

**1. Fix spec misunderstanding.** The original implementation was partly based on a miscommunication of the requirements - that only specific combinations of data source selections could trigger a combined rate calculation. This required tracking a lot of details that turned out to be unnecessary. It's a beautiful thing when the more technically correct way happens to be the simpler way.

**2. Compute rates more directly.** The original implementation performed a series of successive transformations on the input measure data, some of which took place on the frontend. This implementation builds (nearly) display-ready objects straight from the inputs. In some cases this reduces code; fewer transformations are necessary. In other cases it increases code; some arrays with maps have become individual variables in the code.

**3. Separate Rates from Other Things.** Certain measures have, in addition to the typical "numerator, denominator, rate" objects, some non-rate values that still need combining (for example, CPU-AD's count of individuals unreachable for care planning). These values are stored mixed-in with the typical rates, and the original implementation kept them mixed in. Dealing with these values (and their differences from typical rates) will always be complex, but this implementation draws a clear line to hide that complexity away (for the vast majority of measures).

I have left additional comments on the changed files.

### Related ticket(s)
* CMDCT-3651 - Combine rates for admin+admin. Functionality rewritten in this branch.
* CMDCT-3652 - Combine rates for hybrid+hybrid. Functionality rewritten in this branch.
* CMDCT-3653 - Combine rates for hybrid+admin. Functionality rewritten in this branch.
* CMDCT-3877 - Combine values for CPU-AD. Functionality rewritten in this branch.
* CMDCT-3878 - Combine values for PCR-AD. Functionality rewritten in this branch.
* CMDCT-3904 - Use user-overridden rate values when computing weighted combined rate. Bugfix replicated in this branch.
* CMDCT-3903 - Use denominator to weight Hybrid+Admin calculations. Bugfix replicated in this branch.
* CMDCT-3933 - Weighted rate "doesn't recalculate". Bugfix is in this branch only.
* CMDCT-3923 - Display "Not Applicable" for rates that were reported, but cannot be combined. Bugfix is in this branch only.
* CMDCT-3924 - Copy the usable rate to "combined" when one uses an Other data source
* CMDCT-3925 - Copy the usable rate to "combined" when one uses the ECDS data source

---
### How to test
Oh, boy.
1. Log into QMR as a state user, for a state with separated Medicaid & CHIP core sets
2. Fill out some core sets
3. Look at the combined rates.
4. Everything should... work.
5. Least helpful testing instructions ever, I'm sorry.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security


services/app-api/types.ts
benmartin-coforma, line 1: Two big changes in here.

1. The list of possible DataSource values has been moved here, from a file that was specific to combined rates.
2. I did a fair amount of research, across all environments, to enumerate the specific values we can expect for DataSource, DataSourceSelections, DataSourceSelections sub-values, and MeasurementSpecification.
   * There are some exceptions to the new type definitions, where noted in comments. In these cases, I felt it was better to have a type that is convenient for us to use (in 2024 and beyond) than one that is exhaustively correct across all time (but which requires more wrangling & casting).
   * This data was empirically gathered, so while it does represent an exhaustive (except where noted) list of everything we can see today, it may not include everything we _could_ see, and will not automatically include what we _will_ see when options change in future years.
   * I would love to be able to hook this type definition into the data generation process, so that (for example) new options will automatically be included in the type definition. Until then, I still believe these enumerations will be worthwhile, because they will eliminate some of the guesswork of "what does this data actually look like?" and "what edge cases might I be missing"?

services/ui-src/src/hooks/api/useGetRate.tsx
benmartin-coforma, line 1: Let's have a little typing, as a treat.

services/ui-src/src/shared/commonQuestions/AdditionalCombinedValues/AdditionalCombinedValues.tsx
benmartin-coforma, line 1: This is the table of miscellaneous values that will show up beneath the proper N/D/R rates, for CPU-AD and PCR-AD only.

Well technically "beneath" for CPU-AD only, because PCR-AD doesn't have any proper N/D/R rates. CPU-AD is the only measure in the app with both kinds of data.

Note the UID switch in `displayValue()`. This is necessary because we store our rates in the database as numbers. The specific values UIDs listed here are supposed to be displayed with trailing zeros, which a `number` cannot hold, but `.toFixed()` gives us.

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateNDR.tsx
benmartin-coforma, line 1: This was the file where we got some of the the biggest payoffs from the refactor.

1. All of the custom type wrangling and data transformation is gone.
2. All of the logic to switch between N/D/R rates and Other Stuff is gone; the Other Stuff is displayed by `<AdditionalCombinedValues>`.
3. The function which modified rate objects to set default values has been replaced, by a helper function that determines the correct display value on the fly.
4. The separate functions for desktop table / mobile table have been inlined within the main block of markup, because that was the easiest way to dodge some variable scoping issues. The alternative being to pass more arguments to those separate functions, which might have turned out okay; I just went this way first and didn't dislike it enough to go back.
5. The render-time CSS text transformation has been replaced by pre-baked tables of display values, for rate components and program types.

services/ui-src/src/shared/commonQuestions/CombinedRateNDR/CombinedRateTypes.ts
benmartin-coforma, line 1: These types no longer exist; their successors are now defined in ui-src/src/types.ts

services/ui-src/src/shared/commonQuestions/DataSouceInformationBanner/DataSourceInformationBanner.tsx
benmartin-coforma, line 1: This file's diff is almost entirely noise from the type changes. The one intentional change is the new programDisplayNames map... Which you could argue is also just noise, I guess.

services/ui-src/src/shared/commonQuestions/DataSouceInformationBanner/combinedRates.json
benmartin-coforma, line 1: This example payload doesn't need to stick around, now that the feature is developed (and it's out of date anyway).

services/ui-src/src/types.ts
benmartin-coforma, line 1: This type def is 95% copy-paste from the app-api type def.

Someday we'll find a low-hassle way to share types between our services, and on that day we will delete a ton of code. This included.

The other 5% is a simplification; where the typedef in app-api gets very detailed around DataSourceSelections, this one cops out with a `[key: string]` mapped object type.

services/ui-src/src/views/CombinedRatesPage/CombinedRatesMeasure.tsx
benmartin-coforma, line 1: In this file, we finally see the payoff of dropping `.Item` from rates/get in the API. And we also see the three separate components for data source, rates, and additional values rendered side-by-side. It's beautiful.

services/app-api/handlers/rate/dataSourceAnalysis.ts
```diff
@@ -0,0 +1,46 @@
...
+  const DataSource = measure?.data?.DataSource ?? [];
+  const DataSourceSelections = measure?.data?.DataSourceSelections ?? {};
+  const MeasurementSpecification = measure?.data?.MeasurementSpecification;
+
+  const hasOtherDataSource = DataSource.includes(DataSourceTypes.Other);
+  const hasECDSDataSource = DataSource.includes(DataSourceTypes.ECDS);
+  const hasOtherSpecification =
+    MeasurementSpecification === MeasurementSpecificationType.Other;
+  const isUnusableForCalc =
+    hasOtherDataSource || hasECDSDataSource || hasOtherSpecification;
```
BearHanded, line 31: quick question, did you validate the admin other behavior? I know it was an open question

BearHanded:
The deployed branch appears to have changed to rendering the table in all caps, is this intended?
image

ailZhou (commented):
Found a small bug. When you first try to fill out CPU-AD and you don't fill one of the value fields originally, the label doesn't get display. Same issue is happening with PCR-AD. 

Screenshot 2024-08-23 at 11 07 43 AM

And PCR-AD
Screenshot 2024-08-23 at 11 13 24 AM

benmartin-coforma:
> The deployed branch appears to have changed to rendering the table in all caps, is this intended?

Nope! Fixed, I think.

benmartin-coforma:
> Found a small bug. When you first try to fill out CPU-AD and you don't fill one of the value fields originally, the label doesn't get display. Same issue is happening with PCR-AD.

How interesting! I didn't expect that, but it is fixed now.

services/app-api/handlers/rate/dataSourceAnalysis.ts
benmartin-coforma: Um, not explicitly no. But it seems implied by the spec, and that is how it was working in Val during the testing period, and we didn't get any complaints. So I feel pretty confident about it.

services/app-api/handlers/dynamoUtils/scripts/modifyDescription.ts
```diff
@@ -16,39 +15,29 @@ const fetchMeasuresByName = async () => {
...
-    if (
-      foundMeasure.Items &&
-      foundMeasure.Items.length > 0 &&
-      !changeDescriptionTo
-    ) {
-      console.log(
-        `The current description is: ${foundMeasure.Items[0].description}`
-      );
+    if (foundMeasure && foundMeasure.length > 0 && !changeDescriptionTo) {
+      console.log(`The current description is: ${foundMeasure[0].description}`);
```
rocio-desantiago, line 29: Still need this console log? 🤔

services/app-api/handlers/dynamoUtils/scripts/modifyDescription.ts
```diff
@@ -16,39 +15,29 @@ const fetchMeasuresByName = async () => {
...
       changeDescriptionTo = p("What would you like to change it to: ");
       console.log(changeDescriptionTo);
```
rocio-desantiago, line 31: I s'ppose this is the norm for the scripts? To have console.logs

benmartin-coforma: Yes indeed. The console is the only UI for this script - the `p()` function prints a question to the user, which they answer by typing in the terminal. With "the user" being one of us devs, more likely than not.

I'm not sure if we're ever going to need this script in the future... we may not? But I'm inclined to leave it as-is.

benmartin-coforma:
> > The deployed branch appears to have changed to rendering the table in all caps, is this intended?
> 
> Nope! Fixed, I think.

And _now_ it's _really_ fixed.

ailZhou (approved):
approved, looks amazing, you're da best!

ajaitasaini (approved):
reviewed on Thursday / Friday, approving!!

PR #2393, opened 2024-08-22
CMDCT-3732: Eagerly validating api path params for all CRUD handlers

angelaco11:
### Description
Added functions to validate path parameters for various crud functions. Then "eagerly" validating by immediately throwing a bad request error if all path params are not valid when the function is invoked.  

### Related ticket(s)
CMDCT-3732

---
### Pre-review checklist

---
#### Review

#### Security

---


services/app-api/types.ts
```diff
@@ -145,3 +147,31 @@ export interface APIGatewayProxyEvent {
...
+  measure: string | undefined
+) => {
+  return (
+    year &&
+    isValidYear(year) &&
+    !!measures[parseInt(year)].find((measureObject) => {
+      return measureObject.measure == measure;
+    })
+  );
+};
```
benmartin-coforma, line 177: I'd make a couple of tiny changes here.
* This function is currently only called after having already validated `year`, so there's no need to allow undefined values for that parameter
* The isValidYear check is already redundant given the sole call site, but it is still a good idea to check it here. That said, we can check it implicitly by allowing for the key not to exist in `measures`. [Here's an example](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZiEMC8MDeAoGMBGBDAJwC4YBtARgBoYAmagZgF1Ktc8AvEgVzABMBTOAEsw-XiwC+AbgwZQkWEIgA1PABshvAHJcAtjn4EUMABQBrfgE8SFyyDgwolgA7978RNTAkwegwQBKFAA+dFYAenCYX31DCBgACzx4p1cYAAMY-1JGGAAfGB4BYVFedNZ5aGi-OOMEEFJbRhlsSIyAfgA6EWA1LgEIdJgAdyE1NRgCfiguAjBCvkERMRghBygE-hhCAjxLVfiipdKIqIBBPkckqAByeIBzKbwoagNgPC4ILaOSlaV4dRfSydVhTGZzaqxAgQLo9PoDExgAIyaSySogNT8TpqED3EwAIjO+OoSlUGm0NQIBPwBGJtACyLk4AgGKxOLx+IAQnTSepNDoodTCHSAOwMmTozHY3EEgDCPJUfIpgvx+HYdPI4owQA)
* Checking for the truthiness of an object found with `.find` is equivalent to using `.some`
* Triple-eq on the measure abbreviation, because why not 😎 

```suggestion
export const isMeasure = (
  year: string,
  measure: string | undefined
) => {
  return measures[parseInt(year)]?.some((measureObject) =>
      measureObject.measure === measure
  );
};
```

services/app-api/utils/constants/constants.ts
```diff
@@ -20,3 +20,67 @@ export const enum StatusCodes {
...
+  "VT",
+  "VI",
+  "VA",
+  "WA",
+  "WV",
+  "WI",
+  "WY",
+];
+
+export const coreSets = ["ACSM", "ACSC", "CCSM", "CCSC", "ACS", "CCS"];
```
benmartin-coforma, line 86: We do have an existing enumeration of core set abbreviations, in app-api's types.ts. I recommend adding an `isCoreSetAbbr` helper right beneath it:

```typescript
export enum CoreSetAbbr {
  ACS = "ACS", // adult
  ACSM = "ACSM", // adult medicaid
  ACSC = "ACSC", // adult chip
  CCS = "CCS", // child
  CCSM = "CCSM", // child medicaid
  CCSC = "CCSC", // child chip
  HHCS = "HHCS", // health homes
}

export const isCoreSetAbbr = (coreSet: string): coreSet is CoreSetAbbr => {
  return Object.values(CoreSetAbbr).includes(coreSet as CoreSetAbbr);
};
```

My recommendation here is so specific because that will prevent a merge conflict when my combined rates branch lands :D

services/app-api/handlers/measures/create.ts
```diff
@@ -27,10 +36,10 @@ export const createMeasure = handler(async (event, context) => {
     TableName: process.env.measureTableName!,
     Item: {
       compoundKey: dynamoKey,
-      state: event!.pathParameters!.state!,
-      year: parseInt(event!.pathParameters!.year!),
-      coreSet: event!.pathParameters!.coreSet! as CoreSetAbbr,
-      measure: event!.pathParameters!.measure!,
+      state: state,
+      year: parseInt(year),
+      coreSet: coreSet as CoreSetAbbr,
```
benmartin-coforma, line 41: Creating & using the type guard `isCoreSetAbbr` will remove the need for this cast. I believe it's the only such cast in the API.

services/app-api/handlers/measures/get.ts
```diff
@@ -37,7 +44,7 @@ export const listMeasures = handler(async (event, context) => {
 
   let queriedMeasures = await dynamoDb.scanAll<Measure>(params);
   for (let v of queriedMeasures) {
-    const measure = measures[parseInt(year)]?.filter(
+    const measure = measures[parseInt(year as string)]?.filter(
```
benmartin-coforma, line 47: It is very strange that this cast is necessary. Typescript Should Know Better here, and it does know better earlier in the file.

But I don't know what to do about it, beyond what you've already done 🤷

services/app-api/handlers/coreSets/tests/create.test.ts
benmartin-coforma, line 1: Nit: The `list` variable is now unused, on lines 81 and 94.

services/app-api/handlers/measures/get.ts
benmartin-coforma, line 1: This file has the last remaining (easily removable) `pathParameters!`, on line 91.

services/app-api/handlers/measures/tests/create.test.ts
```diff
@@ -56,16 +58,34 @@ describe("Test Create Measure Handler", () => {
...
+
+    expect(res.statusCode).toBe(StatusCodes.BAD_REQUEST);
+    expect(res.body).toContain(Errors.NO_KEY);
+  });
+  test("Fails with bad request when params exist but are not valid", async () => {
+    event.pathParameters = {
+      state: "YA",
+      year: "2020",
+      coreSet: "YLTR",
+      measure: "EEE-EE",
```
benmartin-coforma, line 83: 🐬

services/app-api/handlers/measures/update.ts
```diff
@@ -58,11 +66,14 @@ export const editMeasure = handler(async (event, context) => {
...
     //after updating the database with the latest values for the measure, we run the combine rates calculations for said measure
-    await calculateAndPutRate(
-      event!.pathParameters! as unknown as MeasureParameters
-    ); // TODO, fix this ugly double cast
+    await calculateAndPutRate({
+      state,
+      year,
+      coreSet,
+      measure,
+    });
```
benmartin-coforma, line 76: Love to see a TODO disappear 😎

benmartin-coforma (approved):
This is a great change, and I'd love to merge it in as-is.

I do have a few thoughts for how to take it a bit further, if you have the appetite.
1. It might be nice if `handlers/dynamoUtils/createCompoundKey` accepted an argument of `parameters: CoreSetParameters | MeasureParameters`. MeasureParameters already exists; CoreSetParameters would need to be created right alongside it. I believe that would make both of the `if (...) throw` checks unnecessary, and the four variable initializations would collapse into a single object destructure. It may even be worthwhile to split it into two functions, `createCoreSetKey` and `createMeasureKey`.
2. It would be nice if we could roll in validation for the banner handler parameters. I think that would just be one parameter, the bannerId?
3. I know that you named the validation functions based on my prior suggestion, but I'm wondering if it would be nicer to go a bit more direct. Maybe a quartet of `parseBannerParameters()`, `parseCoreSetParameters()`, `parseMeasureParameters()`, and... um, `parseStateAndYear()`? What do you think?

services/app-api/types.ts
angelaco11: Oh yes this is marvelous. I didn't like that function so I am so glad you came in with an excellent and clean suggestion!!

services/app-api/handlers/coreSets/tests/create.test.ts
angelaco11: nice, good catch!

services/app-api/handlers/measures/create.ts
angelaco11: yup good call, kept that in without thinking about it 😁

services/app-api/handlers/measures/get.ts
angelaco11: great catch. you're the hero of this PR

services/app-api/utils/constants/constants.ts
angelaco11: Yep I love this suggestion. Gives us one source of truth in case it changes

services/app-api/types.ts
angelaco11: updated!

services/app-api/handlers/measures/get.ts
angelaco11: Oh yeah! I added that because I get the following ts error when I run the get.test.ts tests:
Screenshot 2024-08-23 at 10 51 03 AM

I cannot for the life of me figure out why it would claim that it is of type `string | undefined` when it's absolutely never going to be undefined and when hovering over it inline, it's of type string
Screenshot 2024-08-23 at 10 52 43 AM

benmartin-coforma: I dug a little deeper here. It's a shortcoming of the Typescript compiler version we're using in the project. I _think_ it was fixed by the ["improved narrowing" in TS 4.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html#improved-intersection-reduction-union-compatibility-and-narrowing). It seems that the the older version of the compiler has trouble carrying that type information across the block-scope boundary of the for loop.

I'm willing to bet that, if you go to VS Code and `cmd`+`shift`+`p` > `Typescript: Select Typescript Version...` you'll see that you're set to use Visual Studio's version? Which will be version 5.5+. Whereas when our code is compiled, we are currently using version 4.7.3 - one minor version from this fix. If you select the workspace version, and revert your `as string` cast, you'll see the same red squiggle error that you see at test time.

[Typescript Playground demo](https://www.typescriptlang.org/play/?ts=4.7.4#code/MYewdgzgLgBAbgQwDYEsAmCoFMYF4YAUADggE4IC2EAXDAEpailoA80pKYA5gDQzucuAPhgAfGAFcwaLADNOWNAEo8IgN4AoGDFCRYamAE8sZGAF88MEuSowA-HZhqzAbi0wUswsdMAyXzAA9AB6ADpoANQAJIEAdNjQBD6kSiqa2tqkWFASpGBOHhAAasjotFCkEjgIEDrg0HzJ5m7aZu5ZOXkFKMWlaLSyyBDVtbrQzRptGmP6hSWoaI0mpOaWiAuYWAQGybQARABMAAwHACx75kpuYyBIWLFIIFxJy0oagYEZXxnBdhoankIAEIevN0Gl3FAABakEAAdxgYCwCIAoqRYaQCHsACpQzBGZYwOE1DxgdboPZXSbTeq3e6PZ7JN4fb7fX7-WQgFYEO6wFCWI4uDwwFgwACMQv5EXwYoh2hudweTxeZDe2hZrJ+fzaQA) - change the version dropdown in the top left to watch the types change!

benmartin-coforma (approved):

services/app-api/handlers/measures/get.ts
angelaco11: aha!! that makes sense. ok cool i'll add the as string cast back in. I'll also note that doing coercion works here as well (`parseInt(year!)`). Not sure which one you prefer or if it matters much

services/app-api/handlers/measures/get.ts
```diff
@@ -37,8 +44,8 @@ export const listMeasures = handler(async (event, context) => {
 
   let queriedMeasures = await dynamoDb.scanAll<Measure>(params);
   for (let v of queriedMeasures) {
-    const measure = measures[parseInt(year)]?.filter(
-      (m) => m.measure === (v as Measure)?.measure
+    const measure = measures[parseInt(year as string)]?.filter(
+      (m) => m.measure === (v as Measure)?.measurex
```
rocio-desantiago, line 48: is the `measurex` a typo?

rocio-desantiago (approved):
Confirmed in browser path

network-error
ui-error

rocio-desantiago (dismissed):

services/app-api/handlers/measures/get.ts
benmartin-coforma: Nah, doesn't matter one bit.

benmartin-coforma (approved):

ailZhou (approved):
Tested on deploy, found no issues creating, updating and deleting measures & coresets. Thank you for doing the clean up to the handlers! You are free to merge whenever the pr is good 🚀 .

services/app-api/handlers/dynamoUtils/createCompoundKey.ts
angelaco11, line 1: @benmartin-coforma here is the change to update createCompound key to be 2 functions that just create a key based on all params existing. Just to make sure this is in line with your vision

angelaco11:
> This is a great change, and I'd love to merge it in as-is.
> 
> I do have a few thoughts for how to take it a bit further, if you have the appetite.
> 
> 1. It might be nice if `handlers/dynamoUtils/createCompoundKey` accepted an argument of `parameters: CoreSetParameters | MeasureParameters`. MeasureParameters already exists; CoreSetParameters would need to be created right alongside it. I believe that would make both of the `if (...) throw` checks unnecessary, and the four variable initializations would collapse into a single object destructure. It may even be worthwhile to split it into two functions, `createCoreSetKey` and `createMeasureKey`.
> 2. It would be nice if we could roll in validation for the banner handler parameters. I think that would just be one parameter, the bannerId?
> 3. I know that you named the validation functions based on my prior suggestion, but I'm wondering if it would be nicer to go a bit more direct. Maybe a quartet of `parseBannerParameters()`, `parseCoreSetParameters()`, `parseMeasureParameters()`, and... um, `parseStateAndYear()`? What do you think?

These are all excellent ideas. I implemented all of these things and just pushed them up!

benmartin-coforma (approved):
This turned out really nice!

I don't think we need to mock out crateCoresetKey or createMeasureKey, now that they work exclusively on strings (without a dependency on an Event). But those mocks aren't really hurting us, and I don't want to ask for any more changes on this PR.

Thanks for putting it all together 🚀

ailZhou (approved):

PR #2419, opened 2024-09-05
Global Validation Refactor - Simple Consolidation

ailZhou:
### Description

The initial refactoring for global validation which is really just moving files with the exact same code in each year to one shared file between the years.

Files that were made new instead of migrated are :
- TypeValidations.ts - This is just the types files for all fours years merged. Those types file will be moved last so they still have to exist for now
- validatePartialRateCompletion - A couple of files rely on this, but there was slight code different between the years so I did end up having to write some code to combine both.
- dataDrivenTools.ts - This is different between the 4 years so it does have some code to handle that. There still exist a dataDriveTools.ts file for the individual years that needs to exist until all the validations have been migrated

### Related ticket(s)

CMDCT-3959

---
### How to test

If the deploy compiles with no errors, that's a good sign, as the only major errors that would happen is that I did not remember to link the validation call to the shared components. If that was the case, we would have had compile errors and github actions would not deploy.

If you want to test, just pick 1 of the migrated validation and see if the validation pops up

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/shared/globalValidations/validateNoNonZeroNumOrDenom/index.ts
```diff
@@ -1,15 +1,16 @@
+//note: I'm not sure if this is used
```
benmartin-coforma, line 1: Currently:
* `import * as DC from "dataConstants"` - used on lines 80, 116, 117 😛 
* `validateRateZeroOMS` - not used. Shares a common name with a validation function that exists within the globalValidations for each year. Each year's version is identical to each other, but not to the version defined in this file.
* `validateRateNotZeroOMS` - not used. Same deal: identical between years, but this version is different.
* `validateNoNonZeroNumOrDenomPM` - not used. Similar deal, except in prior years it is not global; it's defined only within SSHH.
* The `_`-prefixed functions in this file are referenced only within this file.

services/ui-src/src/shared/globalValidations/validatePartialRateCompletion/index.ts
benmartin-coforma, line 1: I'm not completely following the logic for the error message; my gut tells me it should either break for 2022 & earlier, or for 2023 & later.
* On line 52, we set `multipleCats` to `true` if any categories have a truthy label.
* In 2022-, all measures without multiple categories have an empty array of categories in `rateLabelText`, but I think we dynamically add a special placeholder category at runtime?
* In 2023+, all measures without multiple categories have a single placeholder category, with a unique ID, but empty strings for `label` and `text`.
* So in 2022-, will we erroneously set `multipleCats` to true? Which would result in a strange-looking validation error message for any callers using the default error message function.
   - Which is all of them. No measure ever passes a custom error message function.
* I haven't actually tested this; if error messages from this function _are_ appearing correctly for 2022-, then I guess we're fine here.

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
```diff
@@ -0,0 +1,158 @@
...
+  qualifiers: LabelData[],
+  rateData: RateData
+) => {
+  const rateArray: PM[][] = [];
+
+  for (const cat of categories.map((c) => c.id)) {
+    const tempArr: PM[] = [];
+    for (const qual of qualifiers.map((q) => q.id)) {
+      if (isLegacyLabel()) {
+        tempArr.push(rateData.rates?.[qual]?.[cat]?.[0] ?? {});
```
benmartin-coforma, line 62: Prior to 2023 we had UIDs that were `qual.cat`!? I had no idea x_x

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
```diff
@@ -0,0 +1,158 @@
...
+) => {
+  const dictionary: { [cleanedLabel: string]: string } = {};
+  const checkNode = (node: DDT.SingleOmsNode) => {
+    // dive a layer
+    for (const option of node.options ?? []) {
+      checkNode(option);
+    }
+
+    if (isLegacyLabel()) dictionary[cleanString(node.id)] = node.id;
+    else dictionary[node.id] = node.label!;
```
benmartin-coforma, line 108: Request: I love curly braces

```suggestion
    if (isLegacyLabel()) {
      dictionary[cleanString(node.id)] = node.id;
    } else {
      dictionary[node.id] = node.label!;
    }
```

benmartin-coforma (dismissed):
If you've answered my questions above to your satisfaction, then I say this PR is good to merge 👍

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
ailZhou: Yep and i decided that it was weird and changed it to cat.qual. A part of me wonders everyday if that was the right decision 🫠

services/ui-src/src/shared/globalValidations/validatePartialRateCompletion/index.ts
ailZhou: surprise! this actually works and that's because the category that comes in when its empty for 2022- is actually an empty []. the system does use to add singleCategory as a placeholder label for rates with no categories but i guess its after the validation layer 🤔. It also saving correctly in the db but this does feel like there's some unholy logic going on behind the scenes.

services/ui-src/src/shared/globalValidations/validateNoNonZeroNumOrDenom/index.ts
ailZhou: i'm convinced this is not used or was used at one point and broken down to smaller validations. i'm deleting it

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
ailZhou: added!

PR #2431, opened 2024-09-13
CMDCT-3960: creating new dynamo table for measures with PKs that can be queried instead of scan all

angelaco11:
### Description 
The current dynamo `measures` table has a limitation in its primary key structure. It currently is {state}{year}{coreSet}{measure}. The limitation here is the fact that the measure is tacked on to the primary key which makes each primary key unique. Therefore, when we want to do a query to get all the measures for a state, year and coreset, we cannot do a query on the primary key (since it contains each individual measure) and must do a `scanAll` which is not performant and is causing db threshold alarms. 

The changes here are to write a script that creates a new measures table that is rekeyed. The new primary key is {state}{year}{coreSet} and the range/sort key is the measure name. this allows us to query for all measures for a state, year and coreset since all of those measures will have the same primary key (e.g. all measures from AL, 2024 and CCSM will have the primary key `AL2024CCSM`. The sort/range key will be the measure so that single measures can be retrieved by combining the primary key and the range key. This will make querying the db much more performant and we will not have to use scanAll.

I also updated all calls to query the measures table to point to the new table with the correct keys.

### Related ticket(s)
CMDCT-3960

---
### How to test
https://d2flx34i0vw50w.cloudfront.net/
1. Log in as any user
2. Click on any coreset and make sure that the measures list loads properly

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


services/app-api/handlers/measures/get.ts
```diff
@@ -35,14 +35,14 @@ export const listMeasures = handler(async (event, context) => {
...
-      "list"
-    ),
+    TableName: process.env.measureTable,
+    KeyConditionExpression: "compoundKey = :compoundKey",
+    ExpressionAttributeValues: {
+      ":compoundKey": `${state}${parseInt(year)}${coreSet}`,
+    },
   };
 
-  let queriedMeasures = await dynamoDb.scanAll<Measure>(params);
```
angelaco11, line 45: This is the problem query that sparked the need to rekey the dynamo table. This scanAll is terribly non-performant and has been setting off db threshold alerts such as [these](https://stratiformworkspace.slack.com/archives/C038M9GDA5T/p1725561308236519)

services/database/serverless.yml
```diff
@@ -108,6 +109,25 @@ resources:
           - AttributeName: key
             KeyType: HASH
         BillingMode: PAY_PER_REQUEST # Set the capacity to auto-scale
+    CSMeasureTable:
+      Type: AWS::DynamoDB::Table
```
angelaco11, line 113: I need advice on the naming of this stuff since we can't delete the old table until we are fully done implementing the new one. Let me know if there are better suggestions

services/database/serverless.yml
```diff
@@ -108,6 +109,25 @@ resources:
           - AttributeName: key
             KeyType: HASH
         BillingMode: PAY_PER_REQUEST # Set the capacity to auto-scale
+    CSMeasureTable:
```
angelaco11, line 112: I need advice on the naming of this stuff since we can't delete the old table until we are fully done implementing the new one. Let me know if there are better suggestions

services/app-api/utils/parseParameters.ts
```diff
@@ -29,7 +29,7 @@ export const parseStateAndYearParameters = (event: APIGatewayProxyEvent) => {
     logger.warn("Invalid or missing year in path");
     return { allParamsValid: false as const };
   }
-  return { allParamsValid: true as const, state, year };
+  return { allParamsValid: true as const, state, year: parseInt(year) };
```
BearHanded, line 32: radical

services/database/serverless.yml
BearHanded: Should these all be CoreSetMeasureTable?

services/app-api/handlers/measures/get.ts
```diff
@@ -35,16 +33,16 @@ export const listMeasures = handler(async (event, context) => {
...
+    ExpressionAttributeValues: {
+      ":compoundKey": `${state}${year}${coreSet}`,
+    },
   };
 
-  let queriedMeasures = await dynamoDb.scanAll<Measure>(params);
+  const queriedMeasures = await dynamoDb.queryAll<Measure>(params);
   for (let v of queriedMeasures) {
-    const measure = measures[parseInt(year as string)]?.filter(
+    const measure = measures[year as number]?.filter(
```
benmartin-coforma, line 45: Good news, this cast is no longer necessary!

```suggestion
    const measure = measures[year]?.filter(
```

ajaitasaini (commented):
thank you for the thorough description + all the context provided -- made it so easy to understand why we're making these changes! testing via cloudfront worked as expected and the code makes sense / nothing major is sticking out to me

benmartin-coforma (approved):
Looks good to me!

benmartin-coforma:
@ailZhou Note that we will need to perform the migration in this PR _immediately_ when it merges into each environment, or we risk data loss.

services/database/serverless.yml
benmartin-coforma: I don't have a really strong opinion here... what I would say is, let's not prefix our table name with another valid table name. For that reason I'd vote against `CoreSetMeasure`.

services/app-api/handlers/measures/get.ts
angelaco11: we just chatted about this but for anyone who is curious, typescript is an idiot and unfortunately needs this godforsaken cast even though IT KNOWS year is a mf number.


just a note that if you're looking at this in vscode, it won't throw the error inline, it will only fail during the tests because of a ts mismatch between vscode and qmr. or something

rocio-desantiago (approved):
Code looks good 👍 
Tested with deploy link: Measures showing up for each year as expected ✅

services/database/serverless.yml
angelaco11: renamed to `QualityMeasureTable`

PR #2436, opened 2024-09-16
GV Refactor - omsValidator

ailZhou:
### Description

omsValidator is used to run any validaiton related to the Optional Measure Stratification (OMS) section. So the omsValidator.tsx script has to handle checking the OMS N/D/R fields and act as a gateway to call other OMS related components that get passed in as a callback function. 

This code is a full refactor as the previous code was hard to merge across every year. The full refactor could still be refactored better (espcially the `getOMSRates` function),  so if you have any better ideas of how to make the code even more readable, please let me know!

**So What Do These New Code Do**
- `errorToFillNDR` & `errorForNDRSelection`: These functions were added to cut down on having to write the same error message several times. Not necessary but it does reduce repetitive code.
-  `getOMSRates`: This is the meat of the whole component. The function goes through all the checkboxes and retrieves the rate information that is needed to be checked against the validator. I really want to refactor this to be smaller but hopefully it is more readable than before. 
- `validateNDR`, `validateFields` & `validateValues`: These three functions are the different types of checks we use to validate the data in the rates field. 
     - `validateNDR` is used to check our typical N/D/R sets
     - `validateFields` is used for more unique ones that don't have N/D/R sets, like AIF-HH & IU-HH.  
     - `validateValues` is used for PCR rates
- `buildOPMLocationDictionary`: This builds a label lookup for Other Performance Measure's OPM section. It matches the structure they have set for looking up OMS labels.
- `omsValidations`: This is how the other scripts will call this script. The order of operations are:
    - Initial setup by checking if OMS data is OPM - OMS data, build the opm label dictionary and get the rate data and storing it in `omsRates`
    - Loop through omsRates and check each rates, 
        - if rateData exist, we run it against specific validate functions  and get those error labels to store in `errorArray` 
        - if rateData does not exist, create and store error in `errorArray` variable
        - Then we run any callback functions that are being passed in
        - Then we run `validatePartialRateCompletionOMS`, this I had kept from the previous code because I didn't understand why it had to be structured like this so I choose not to adjust it.
    - After we look through omsRates, we should have all the error messages needed and then we return it back to the measure validation to be rendered

**Preemptively Answering Questions You May Have**
_What happened to the checkIsFilled prop?_ 
I am still confused about the existence of this prop, from what I could see, it was always true so it was never used. I guess the idea was to allow the callbacks to run but turn off n/d/r validation checks? I honestly think that we should refactor this more so that we wouldn't be running other OMS validation through here as if it was a highway but that's a battle for latter.

_Why was there an addition to _validateOPMRates_ script?_
That chunk of code that I added to validateOPMRates was originally in the omsValidator, I thought it made more sense to move it.

_In the omsValidator unit test, it looks like you just changed the expected return of errors.length?  Are you just trying to avoid writing a real test?_ 
So my changes does add more validation messages back but they are also more **accurate**. It was not consistent on when you would see the validation message and it was not generating a message for each qualifier in a category and only generate for the first one. There is a chance that the numbers is still wrong but its not any less wrong than before 😫.

If this is still all too confusing and you want a walkthrough, please let me know!

### Related ticket(s)

CMDCT-3965

---
### How to test

1) Sign into QMR, any state user
2) In Reporting Year 2024, select an adult core set
3) Enter AAB-AD
4) In the Measure Specification Section, select the first radio button
5) Scroll to the Performance Measure Section, and fill our all the N/D/R sets
6) Now Screen down to the the OMS section
Here is what the validation is at different stages of selection:

Screenshot 2024-09-25 at 10 05 08 AM

Screenshot 2024-09-25 at 10 05 14 AM
---

Screenshot 2024-09-25 at 10 05 46 AM

Screenshot 2024-09-25 at 10 05 52 AM
---

Screenshot 2024-09-25 at 10 06 39 AM

Screenshot 2024-09-25 at 10 06 49 AM
---

Screenshot 2024-09-25 at 10 07 31 AM

Screenshot 2024-09-25 at 10 07 39 AM

### Notes

N/A

---
### Pre-review checklist

---
#### Review

#### Security

---


rocio-desantiago (approved):
Tested in deploy link and works as expected from instructions. 
The file `services/ui-src/src/shared/globalValidations/omsValidator/index.ts` could be a good one to huddle on as a team as it's a bit complex for my tiny brain 😬

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
```diff
@@ -123,11 +123,19 @@ export const omsLocationDictionary = (
...
     labels.reduce((prevValue, currentValue, index) => {
       if (index === 0) {
         return dictionary[currentValue] ?? currentValue;
       }
+      if (prevValue === "") {
+        return `${dictionary[currentValue] ?? currentValue}`;
+      }
+      if (dictionary[currentValue] === "") {
+        return prevValue;
+      }
```
BearHanded, line 138: Can this section be simplified? The first two conditionals have the same return

services/ui-src/src/shared/globalValidations/omsValidator/index.ts
```diff
@@ -0,0 +1,306 @@
...
+        .flatMap((additional) =>
+          additional.additionalSubCategories?.map((subCat) => ({
+            key: `${locationDictionary([topLevelKey])} - ${
+              additional.description
+            } - ${subCat.description}`,
+            rateData: subCat.rateData,
+          }))
+        );
+
+      if (additionalSubCategories) {
```
BearHanded, line 129: same here as far as always truthy

services/ui-src/src/shared/globalValidations/dataDrivenTools.ts
ailZhou: oh wow, great catch. that's a big miss on my end

services/ui-src/src/shared/globalValidations/omsValidator/index.ts
```diff
@@ -0,0 +1,306 @@
...
+      omsRates.push({ key: locationDictionary([topLevelKey]), ...topLevel });
+    }
+
+    //if user choose to [+Add Another Classification]
+    if (topLevel.additionalSelections) {
+      const additional = topLevel.additionalSelections?.map((selection) => ({
+        key: `${locationDictionary([topLevelKey])} - ${selection.description}`,
+        rateData: selection.rateData,
+      }));
+      if (additional) {
```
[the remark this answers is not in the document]
ailZhou: thank you for catching this! updated the code to just push directly on the map.

services/ui-src/src/shared/globalValidations/omsValidator/index.ts
ailZhou: thank you for catching this! updated the code to just push directly on the map.

benmartin-coforma (approved):
Yay!

BearHanded (approved):

PR #2454, opened 2024-09-28
[QMR] CMDCT-3856 Remove browserify

rocio-desantiago:
### Description
@benmartin-coforma confirmed that while previously we had discovered `browserify` was a dependency for cypress tests, that removing them shouldn't affect cypress from working. It turns out, the tests are still passing after successfully removing the browserify package and devDependencies. As long as we don't need `cypress-tags`, we should be good to push this update.

### Related ticket(s)

CMDCT-3856

---
### How to test

- [x] Cypress End-to-End Tests passing
- [ ] After merged in, confirm that running `yarn install` doesn't magically add `browserify` back

### Notes

commands ran to confirm browserify removed
  - Confirming dependencies so that `yarn install` doesn't re-install `browserify`
`yarn why browserify`
`yarn why cypress-tags`
`yarn remove cypress-tags`
`yarn remove browserify`
  - To confirm no lingering dependencies
`grep browserify yarn.lock`
  - Clean up `node_modules` and ensure consistency across `package.json` and `yarn.lock` files
`yarn install --check-files`

### DIFF
--- a/services/ui-src/public/browserconfig.xml
+++ b/services/ui-src/public/browserconfig.xml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<browserconfig>
-    <msapplication>
-        <tile>
-            <square150x150logo src="/mstile-150x150.png"/>
-            <TileColor>#da532c</TileColor>
-        </tile>
-    </msapplication>
-</browserconfig>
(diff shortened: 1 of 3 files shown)


ailZhou (approved):
git cypress test ran and passed, so that's good enough for me!

BearHanded (approved):

PR #2459, opened 2024-10-04
CMDCT-4029: rekey rates table

angelaco11:
### Description
rekeying rate table to match the primary key convention of the new coreset and measure tables. We didn't really need to do this but it's nice to be on the same page as the other db tables

### Related ticket(s)

CMDCT-3069, CMDCT-4019, CMDCT-4029

---
### How to test
General smoke testing involving combined rates

### DIFF
--- a/services/uploads/src/dynamoSync/handler.js
+++ b/services/uploads/src/dynamoSync/handler.js
@@ -27,9 +27,37 @@ const uploadFileToS3 = async (filePath, scanResult) => {
 
 const syncDynamoToS3 = handler(async (_event, _context) => {
   console.log("Syncing Dynamo to Uploads");
-  const measureResults = await scanAll(process.env.measureTable);
-  const coreSetResults = await scanAll(process.env.coreSetTable);
-  const rateResults = await scanAll(process.env.rateTableName);
+  let measureResults = await scanAll(process.env.measureTable);
+  let coreSetResults = await scanAll(process.env.coreSetTable);
+  let rateResults = await scanAll(process.env.rateTable);
+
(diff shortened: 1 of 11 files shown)


services/uploads/src/dynamoSync/handler.js
angelaco11, line 1: @BearHanded here is the change we are doing for mathematica to copy over the old compound keys. Let me know if you have design feedback. I decided to just reassign the variables instead of making new ones so that in the future we can just remove that chunk and not have to change anything else

BearHanded (approved):

services/uploads/src/dynamoSync/handler.js
BearHanded: reassigning keeps it stable for them, good call

ailZhou (commented):
So I tested it on the deploy site, let me know if it's a local test only, because I'm getting saving errors. I also have a screenshot of the values I've entered in if you want it.

Screenshot 2024-10-08 at 9 01 10 AM

**Console Error Log**
Screenshot 2024-10-08 at 9 02 01 AM

angelaco11:
> So I tested it on the deploy site, let me know if it's a local test only, because I'm getting saving errors. I also have a screenshot of the values I've entered in if you want it.

Good catch Aileen! I messed up the db query and then didn't test editing a measure 😭 

This should now be resolved!

ailZhou (approved):
lgtm! tested in deploy and the saving issue is gone

benmartin-coforma (approved):

rocio-desantiago (approved):

PR #2462, opened 2024-10-07
[CMDCT-2535] S3 Integration Proposal

BearHanded:
### Related ticket(s)

The current mathematica integration runs every 30 minutes across all data. It takes that data and stores it in an s3 bucket forever. This PR does two things:

1. Sets up a high and low frequency mode for the lambda. This should make it obvious what to switch when higher rates are needed for the QMR testing period (June-Sept), and provides a single value to flip. I debated allowing any cron but it felt less obvious what was going on.
2. Adds a cleanup process to the lambda. After running, the lambda will check the bucket files and remove any older than 1 week. We can shift this if we want. This only will chunk 1000 max for deletion each run, but over time that should vastly outpace the new file rate.

---
### How to test

I can walk through testing with anyone who wants to try it.
- Add a file with a timestamp outside the range to the s3 bucket
- invoke the lambda manually
- the file should be deleted

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/uploads/src/dynamoSync/handler.js
+++ b/services/uploads/src/dynamoSync/handler.js
@@ -32,7 +32,7 @@ const uploadFileToS3 = async (client, filePath, scanResult) => {
 const cleanupFolders = async (client) => {
   const paths = ["CSVmeasures", "CSVcoreSet", "CSVrate"];
 
-  for (const folder in paths) {
+  for (const folder of paths) {
     const path = `coreSetData/${folder}/`;
     const params = {
       Bucket: process.env.dynamoSnapshotS3BucketName,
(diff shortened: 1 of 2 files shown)


ailZhou (approved):
lgtm!

services/uploads/src/dynamoSync/handler.js
```diff
@@ -25,6 +31,38 @@ const uploadFileToS3 = async (filePath, scanResult) => {
...
+      MaxKeys: 1000, // Limited by 1000 per delete
+    };
+    const response = await client.send(new ListObjectsV2Command(params));
+    const files = response.Contents.map((file) => file.Key);
+
+    const cutOffDate = new Date();
+    cutOffDate.setDate(cutOffDate.getDate() - 7);
+
+    const outdated = files.filter((file) => {
+      const dateString = file.split("/").pop().slice(0, -4);
```
benmartin-coforma, line 51: Thought: This would be a little bit clearer as `baseName(file, ".csv")`, but you would also need to `import { baseName } from "node:path"`.

Alternatively, `.at(-1)` may or may not be clearer than `.pop()`

But this is also fine as-is 👍

benmartin-coforma (approved):

PR #2465, opened 2024-10-16
CMDCT-3883: Adding a featuresByYear that will replace year checks around the application

angelaco11:
### Description
Instead of checking what year it is and rendering something for that year like:
```
if (year >= 2024) { 
// render something specific to years 2024 and beyond
}
```
we want to create a `featuresByYear` file that has a bunch of functions. These functions do the same exact year check, except the value add is that they document the feature. So for example for showing the combined rates button for years 2024+, instead of doing that year check (if year >= 2024, render combined rates button), we have a function in the featuresByYear folder that is called `hasCombinedRates` or something. Then that function returns year >= 2024, but when we call it in app, we'll call `featuresByYear.hasCombinedRates` which will document the feature and will make it easier to read and figure out what the code is doing. 

I'm very VERY open to any suggestions here on how to improve this solution. I don't really like what i'm doing here, I think it's clunky and weird and I had tons of trouble actually coming up with useful and descriptive names for a lot of the features. So open to more descriptive names or just another way to do it in general. 

Also, I know this is a pain in the ass to review so let me know if a walk through of the PR may be more helpful.

### Related ticket(s)
CMDCT-3883

---
### How to test
https://d3oh08q9m9o734.cloudfront.net/
This is a pretty sweeping PR so I would think a general smoke test is in order here. Just make sure things all look as they should basically.

### DIFF
--- a/services/ui-src/src/components/Table/columns/measures.tsx
+++ b/services/ui-src/src/components/Table/columns/measures.tsx
@@ -84,7 +84,7 @@ export const measuresColumns = (
         );
       },
     },
-    ...(featuresByYear.displayMandatoryColumn
+    ...(featuresByYear.displayMandatoryMeasuresColumn
       ? [
           {
             header: "Mandatory",
--- a/services/ui-src/src/components/Table/columns/measures.tsx
+++ b/services/ui-src/src/components/Table/columns/measures.tsx
@@ -4,6 +4,7 @@ import { BsCheck } from "react-icons/bs";
 import { Link } from "react-router-dom";
 import { format } from "date-fns";
 import { MeasureTableItem, TableColumn } from "../types";
+import { featuresByYear } from "utils/featuresByYear";
 
 // Get status string from measure data
 const getStatus = (data: MeasureTableItem.Data): MeasureTableItem.Status => {
@@ -55,7 +56,6 @@ const MeasureStatusText = ({
 export const measuresColumns = (
   year: string
 ): TableColumn<MeasureTableItem.Data>[] => {
-  const displayMandatoryColumn = year >= "2024";
   return [
     {
       header: "Abbreviation",
@@ -84,7 +84,7 @@ export const measuresColumns = (
         );
       },
     },
-    ...(displayMandatoryColumn
+    ...(featuresByYear.displayMandatoryColumn
       ? [
           {
             header: "Mandatory",
--- a/services/ui-src/src/shared/coreSetByYear.tsx
+++ b/services/ui-src/src/shared/coreSetByYear.tsx
@@ -1,5 +1,6 @@
 import { SPA } from "libs/spaLib";
 import { AnyObject, CoreSetAbbr } from "types";
+import { featuresByYear } from "utils/featuresByYear";
 
 export type CoreSetType = "coreSet" | "text";
 
@@ -34,12 +35,12 @@ export const coreSetType = (abbr: string) => {
   return;
 };
 
-export const coreSetSubTitles = (year: string, abbr: string) => {
+export const coreSetSubTitles = (abbr: string) => {
   let lastChar = abbr[abbr.length - 1];
   let coreType = coreSetType(abbr) || "";
   let list: AnyObject = {};
   //using the last char of the abbr, we can determine if there's a subtitle
-  if (parseInt(year) <= 2023) {
+  if (!featuresByYear.uniqueCoreSetTitles) {
     list = {
       C: "Chip",
       M: "Medicaid",
@@ -60,18 +61,18 @@ export const coreSetSubTitles = (year: string, abbr: string) => {
   return list[lastChar] || "";
 };
 
-export const coreSetTitles = (year: string, abbr: string, type?: string) => {
-  const subTitle = coreSetSubTitles(year, abbr);
+export const coreSetTitles = (abbr: string, type?: string) => {
+  const subTitle = coreSetSubTitles(abbr);
   const subType = type || "Measures";
   let name = `${coreSetType(abbr)} Core Set ${subType}`;
   return subTitle ? `${name}: ${subTitle}` : name;
 };
 
 //seperated coresets have unique titles for their breadcrumb menu. this is only for 2024 and onward
-export const coreSetBreadCrumbTitle = (
-  year: string
-): { [key: string]: string } | undefined => {
-  if (parseInt(year) >= 2024)
+export const coreSetBreadCrumbTitle = ():
+  | { [key: string]: string }
+  | undefined => {
+  if (featuresByYear.uniqueCoreSetTitles)
     return {
       [CoreSetAbbr.ACSC]: "(Separate CHIP)",
       [CoreSetAbbr.ACSM]: "(Medicaid (Title XIX & XXI))",
--- a/services/ui-src/src/shared/coreSetByYear.tsx
+++ b/services/ui-src/src/shared/coreSetByYear.tsx
@@ -40,7 +40,7 @@ export const coreSetSubTitles = (abbr: string) => {
   let coreType = coreSetType(abbr) || "";
   let list: AnyObject = {};
   //using the last char of the abbr, we can determine if there's a subtitle
-  if (!featuresByYear.uniqueCoreSetTitles) {
+  if (!featuresByYear.hasCombinedRates) {
     list = {
       C: "Chip",
       M: "Medicaid",
@@ -72,7 +72,7 @@ export const coreSetTitles = (abbr: string, type?: string) => {
 export const coreSetBreadCrumbTitle = ():
   | { [key: string]: string }
   | undefined => {
-  if (featuresByYear.uniqueCoreSetTitles)
+  if (featuresByYear.hasCombinedRates)
     return {
       [CoreSetAbbr.ACSC]: "(Separate CHIP)",
       [CoreSetAbbr.ACSM]: "(Medicaid (Title XIX & XXI))",
--- a/services/ui-src/src/shared/coreSetByYear.tsx
+++ b/services/ui-src/src/shared/coreSetByYear.tsx
@@ -1,6 +1,7 @@
 import { SPA } from "libs/spaLib";
-import { AnyObject, CoreSetAbbr } from "types";
+import { CoreSetAbbr } from "types";
 import { featuresByYear } from "utils/featuresByYear";
+import { assertExhaustive } from "utils/typing";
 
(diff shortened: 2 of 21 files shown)


services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -0,0 +1,37 @@
...
+  get pheYear() {
+    return getMeasureYear() < 2023;
+  },
+  get optionalDateRangeTracking() {
+    return getMeasureYear() >= 2023;
+  },
+  get uniqueCoreSetTitles() {
+    return getMeasureYear() >= 2024;
+  },
+  get yearIsNot2025() {
```
BearHanded, line 22: this one feels odd to me. is this a companion to a feature flag somewhere?

services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -0,0 +1,37 @@
...
+    return getMeasureYear() !== 2025;
+  },
+  get displayMandatoryColumn() {
+    return getMeasureYear() >= 2024;
+  },
+  get humanReadableTextForMathematica() {
```
BearHanded, line 28: I think maybe a descriptive label for what this one is doing, with comments here saying it is for mathematica's benefit? I think it adds category labels for them?

services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -0,0 +1,37 @@
...
+  get optionalDateRangeTracking() {
+    return getMeasureYear() >= 2023;
+  },
+  get uniqueCoreSetTitles() {
+    return getMeasureYear() >= 2024;
+  },
+  get yearIsNot2025() {
+    return getMeasureYear() !== 2025;
+  },
+  get displayMandatoryColumn() {
```
BearHanded, line 25: displayMeasuresMandatoryColumn?

services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -0,0 +1,37 @@
...
+  get hasCombinedRates() {
+    return getMeasureYear() >= 2024;
+  },
+  get hasQualCatLabels() {
+    return getMeasureYear() <= 2022;
+  },
+  get hasMandatoryReporting() {
+    return getMeasureYear() >= 2024;
+  },
+  get pheYear() {
```
BearHanded, line 13: potentialPheYear? Not sure on this one, I just know this is combined with a flag to determine the message

services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -0,0 +1,37 @@
...
+  get yearIsNot2025() {
+    return getMeasureYear() !== 2025;
+  },
+  get displayMandatoryColumn() {
+    return getMeasureYear() >= 2024;
+  },
+  get humanReadableTextForMathematica() {
+    return getMeasureYear() >= 2023;
+  },
+  get displayCheckmarkWarning() {
```
BearHanded, line 31: what checkmark?

BearHanded (commented):
I like where this is headed. I think some of the names can be a bit more descriptive about what is controlled, potentially with a comment in the new file adding more background if needed.

services/ui-src/src/shared/coreSetByYear.tsx
```diff
@@ -60,18 +61,18 @@ export const coreSetSubTitles = (year: string, abbr: string) => {
...
 
 //seperated coresets have unique titles for their breadcrumb menu. this is only for 2024 and onward
-export const coreSetBreadCrumbTitle = (
-  year: string
-): { [key: string]: string } | undefined => {
-  if (parseInt(year) >= 2024)
+export const coreSetBreadCrumbTitle = ():
+  | { [key: string]: string }
+  | undefined => {
+  if (featuresByYear.uniqueCoreSetTitles)
```
benmartin-coforma, line 75: I think we might be better off merging this flag with `hasCombinedRates`; the reason we changed the breadcrumb is tightly tied to the existence of the newly-separated core sets. Maybe something like this:

```
  /**
   * Prior to 2024, Adult Core Sets were always reported combined.
   * In 2024, we separated Adult-CHIP from Adult-Medicaid, for some states.
   * At the same time, we added the Combined Rates feature,
   * to provide a summary view for data across the new separation.
   *
   * This goes hand-in-hand with other small features,
   * like more specific core set titles in the header and breadcrumbs.
   */
  get hasCombinedRates() {
    return getMeasureYear() >= 2024;
  },
```

benmartin-coforma: I also think this function (`coreSetBreadCrumbTitle()`) might be due for a rewrite, but it will be a little bit more involved because we'll also have to change the call sites. I won't ask for it as a part of this PR.

benmartin-coforma (commented):
Hey, this is looking nice! My review is still in progress, but I'm posting this today so you can see my comments so far.

I'm commenting on the code where the flags are used rather than where they're declared, just so that featuresByYear.ts doesn't end up with a million comments on it.

benmartin-coforma (commented):
Again, looking good so far.

In addition to my comments below, I have three notes.
1. I'm not sure what to do about the year checks in MonthPicker/calendarPopup.tsx. I suspect they're due to the discrepancy between `HEDIS year` and `HEDIS MY year`. In 2021 we stopped using `HEDIS year`, and the reporting year dropdown goes back three years. So in 2023 we were able to remove some references to `HEDIS year`; compare the common question label MeasureDescriptor between 2022 and 2023 to see the difference. If the logic in calendarPopup.tsx is there for the same reason, we may want a feature named `includeHedisYearExplanation` or some such. For more details we've got Aileen, and CMDCT-2763.
2. The hardcoded "2020" in deliverySystems.tsx is still a mystery to me.
3. We did rework the OMS categories in 2023; it may be worth a flag with a brief explanation of the what & why for that. See https://jiraent.cms.gov/browse/CMDCT-2442 (and related epic & stories) for details there. Here's my stab at a possible flag & comment:
```
  /**
   * Prior to 2023, the OMS section included every rate from the PM section,
   * allowed for subcategorization of every category, and included categories
   * such as "primary language" and "disability status".
   *
   * In 2023 we excluded some rates, disabled some custom subcategorizations,
   * and removed several categories and sub-categories entirely.
   */
  get hasStreamlinedOms() {
    return getMeasureYear() >= 2023
  }
```

services/ui-src/src/components/Table/columns/measures.tsx
```diff
@@ -84,7 +84,7 @@ export const measuresColumns = (
         );
       },
     },
-    ...(displayMandatoryColumn
+    ...(featuresByYear.displayMandatoryColumn
```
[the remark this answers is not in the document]
benmartin-coforma: "we added a the" great comment good work ben

services/ui-src/src/utils/getLabelText.tsx
```diff
@@ -42,8 +43,5 @@ export const getLabelText = (): { [key: string]: string } => {
 
 //pre-2023, the system was using string types for categories & qualifiers. we want to be able to identify it as it determines our data structure for those years
 export const isLegacyLabel = () => {
-  const { pathname } = window.location;
-  const params = pathname.split("/");
-  const year = Number(params[2]);
-  return year < 2023;
+  return featuresByYear.hasQualCatLabels;
```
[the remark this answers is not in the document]
benmartin-coforma: @ailZhou , is that comment a good summary of the Great Label Change of 2023?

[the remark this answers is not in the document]
ailZhou: yeah i think this is a good summary of what happened.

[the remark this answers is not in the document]
ailZhou: My vote is on not ripping out isLegacyLabel, atm. Definitely a future refactor PR, probably a little later into QMR refactoring like after we figure out how to truncate the measures somehow 🤔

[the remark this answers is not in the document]
benmartin-coforma: Sure, that's fine with me.

---
Just to clarify, though: when I say there's no need for isLegacyLabel, I just mean that every instance of
```
if (isLegacyLabel()) { ... }
```
could (in theory) now be replaced with
```
if(featuresByYear.hasQualCatLabels) { ... }
```
which is something that we _should_ do sooner than later, imo, because `QualCatLabel` is a more informative name than `LegacyLabel` &mdash; especially with a comment like this one available on mouseover. It would have saved me a bit of time spent confused while code reviewing validation PRs 😅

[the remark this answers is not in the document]
angelaco11: yeah we could go around and replace all the isLegacyLabel() checks with the qualcatlabel thing. Do we want to do that in this PR? There's a LOT of isLegacyLabel() checks. I would be cool doing a separate PR after this gets merged just for that one lol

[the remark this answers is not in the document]
benmartin-coforma: Yeah, let's do that later.

services/ui-src/src/shared/coreSetByYear.tsx
```diff
@@ -36,29 +37,40 @@ export const coreSetType = (abbr: string) => {
...
+      case CoreSetAbbr.CCS:
+      case CoreSetAbbr.CCSM:
+        return "Medicaid (Title XIX & XXI)";
+      case CoreSetAbbr.ACSC:
+      case CoreSetAbbr.CCSC:
+        return "Separate CHIP";
+      case CoreSetAbbr.HHCS:
+        return "";
+      default:
+        assertExhaustive(abbr as never);
```
benmartin-coforma, line 53: The `as never` cast here defeats the purpose of `assertExhaustive()` &mdash; we want Typescript to confirm for us that we have covered every possible value.

This is possible for Typescript to do when `abbr` is one of a limited set of values (that is, an enum). It is impossible for Typescript to do when `abbr` could be any string.

So, to make `assertExhaustive` worthwhile, we will need to change the type signature of the function, and then chase any type errors up the call stack, adding checks and/or casts as needed. Otherwise, we might as well leave `assertExhaustive` out.

benmartin-coforma (approved):
I love this. Thanks for putting up with all of my nitpicks; let's get this merged!

benmartin-coforma (approved):

rocio-desantiago (approved):
Smoke tested years 2021-2025 and all working as usual 👍

PR #2477, opened 2024-11-05
CMDCT-4034: Removing FFY language for 2025

angelaco11:
### Description
Created one of those featureByYear flag things that will not display "FFY" string when the year is 2025 or beyond. 

### Related ticket(s)
CMDCT-4034

---
### How to test
https://djh52euni05mh.cloudfront.net/
1. Go around 2025 and make sure there is no mention of `FFY 2025` (you will see some instances of `FFY 2024` and that is ok and expected)
2. Make sure all other years do have FFY displayed

### DIFF
--- a/services/ui-src/src/utils/featuresByYear.ts
+++ b/services/ui-src/src/utils/featuresByYear.ts
@@ -100,4 +100,12 @@ export const featuresByYear = {
   get hasStreamlinedOms() {
     return getMeasureYear() >= 2023;
   },
+  /**
+   * Prior to 2025, we display FFY before the year name.
+   *
+   * In 2025 and beyond, we will not display the FFY language in front of the year.
+   */
+  get displayFFYLanguage() {
+    return getMeasureYear() < 2025;
+  },
 };
--- a/services/ui-src/src/utils/featuresByYear.ts
+++ b/services/ui-src/src/utils/featuresByYear.ts
@@ -101,9 +101,10 @@ export const featuresByYear = {
     return getMeasureYear() >= 2023;
   },
   /**
-   * Prior to 2025, we display FFY before the year name.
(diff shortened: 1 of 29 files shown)


BearHanded (dismissed):
nice featuresByYear thing

services/ui-src/src/utils/featuresByYear.ts
```diff
@@ -100,4 +100,12 @@ export const featuresByYear = {
   get hasStreamlinedOms() {
     return getMeasureYear() >= 2023;
   },
+  /**
+   * Prior to 2025, we display FFY before the year name.
+   *
+   * In 2025 and beyond, we will not display the FFY language in front of the year.
+   */
```
benmartin-coforma, line 107: I'm going to be really nitpicky about the comment and request some more "why". Something like this:

```suggestion
  /**
   * Prior to 2025, we displayed FFY before the year name. This was done in part
   * to differentiate Federal Fiscal Year and HEDIS Medical Year.
   *
   * In 2025, stakeholders decided this explicit differentiation was unneeded.
   */
```

(and it isn't really _super_ important, but this comment would also bring the max line length back down to 80 chars)

benmartin-coforma (approved):
This covers every instance of "FFY" I can find in the app, except for three:
1. In the Support Info page, we have "Beginning with FFY 2024 reporting, states are required to report..."
2. In the coreSetByYear component, we have "Only one group of Adult Core Set Measures can be submitted per FFY" (three times)
3. On the Combined Rates page, we have "...states are not asked to report performance measure data for these measures for FFY 2024 Core Set Reporting in the online reporting system: CPC-CH, LBW-CH..."

If you're already aware of these exceptions and we intend for them to stay, then I think we're all good!

benmartin-coforma (approved):

ailZhou (approved):

PR #2503, opened 2024-11-25
CMDCT-3843 Upgrading react-query (and other packages)

peoplespete:
### Description
This is the massive PR for upgrading react-query. It got complicated by the fact that to upgrade react-query we had to upgrade react and then react-dom and then ....and then .... You get the picture. I finally got things to a stable place where react-query was upgraded and all tests are working. There are a couple follow up tickets detailed below to mop up some of the visual or console issues this ticket's code will create.

### Related ticket(s)
CMDCT-4178 - this is a ticket to get the look and feel back after this update (which caused some visual ugliness)
CMDCT-4179 - this is a ticket to clean up some console errors left by this update (they are warnings not a sign of something actually broken)

### How to test
Pull the branch and yarn install at root and at `/service/ui-src/` to get all packages up to date.
Then `./run local` and then poke around to see that the application still works.
You can also run this code here: https://d1dh9p0hgqbcd5.cloudfront.net/

### Notes
This is where the most important changes came:
https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2503/files#diff-3709c9a794263a6b304bf1e61899bd231873695fc6931c550b57757bb3d8ad46
From there major changes had to do with the way chakra does toasts and the way react createRoot works now. Also chakra changed their full width property. There were also a few changes related to the upgrade of cms's design package.
Then there are a slew of files that have an identical change which is to move a single line into a wait block.

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/package.json
```diff
@@ -3,11 +3,15 @@
...
   "dependencies": {
-    "@chakra-ui/react": "^1.8.9",
+    "@chakra-ui/react": "2.0.0",
+    "@chakra-ui/toast": "2.0.0",
+    "@chakra-ui/styled-system": "^2",
     "@cmsgov/design-system": "^5.0.2",
     "@emotion/react": "^11.13.3",
     "@emotion/styled": "^11.13.0",
     "@hookform/resolvers": "^2.9.11",
+    "@tanstack/react-query": "^5",
```
peoplespete, line 13: Upgrading this one required changes to chakra and react and related packages.

services/ui-src/setupJest.ts
```diff
@@ -3,6 +3,20 @@ import "@testing-library/jest-dom";
 import { mockFlags, resetLDMocks } from "jest-launchdarkly-mock";
 
 global.React = React;
+Object.defineProperty(window, "matchMedia", {
```
peoplespete, line 6: This solved a testing issue that came up related to different versions of packages.

services/ui-src/src/measures/2021/SSHH/index.test.tsx
benmartin-coforma, line 1: I'd like to note that, while I believe this code will work, it's against a few Testing Library best practices. I have a lot of respect for Testing Library; it's very well-considered and is one of the few 3rd-party dependencies I would willingly take on.

One is that, generally, a `waitFor` callback [should contain only one assertion](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback). This doesn't affect correctness but it can affect performance in failure cases. Not a big deal.

Another is that, usually, `query` variants [should only be used to assert that an element is _absent_](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence). Again, correctness is not affected; the effect is just a worse error message if the element can't be found. I understand that's how the code was when you found it, and I'm not asking you to change it - I just wanted to note it while I'm in here.

Another is that, generally, `*ByTestId` is [less preferable than `*ByRole` or `*ByText`](https://testing-library.com/docs/queries/about/#priority) - only because that means our tests are interacting with the page more like users would (since users can't see data-test-id attributes). Obviously changing this would be out of scope for the PR; I'm just noting it while we're here.

Finally, there's a guideline to [prefer `find*` over `waitFor`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find). I'm not sure if this will be applicable here, because to be honest I'm not sure why the `waitFor` is here to begin with. As far as I understand it, these two expectations _should_ have worked fine without a `waitFor` at all. I'm probably wrong, because whoever wrote these tests probably didn't add all the extra waiting code for fun, but you never know.

---

The upshot of that wall of text is: will these `expect`s work outside of a `waitFor`? And if not, what is the problem? It's almost certainly out of scope for this PR to address, but if nothing else I'd like to put in a tech debt ticket to clean up these tests later.

services/ui-src/src/measures/2025/AIFHH/index.test.tsx
```diff
@@ -73,9 +73,6 @@ describe(`Test FFY ${year} ${measureAbbr}`, () => {
       </Suspense>
     );
   });
-  afterEach(() => {
-    screen.debug();
-  });
```
benmartin-coforma, line 78: Nice, thank you!

services/ui-src/src/measures/2021/rateLabelText.ts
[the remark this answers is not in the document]
peoplespete: Not exactly sure how this happened. Backed it out so we don't have the noise here. Maybe someday we should all get on the exact same prettier page but this is not the place for that :)

services/ui-src/package.json
```diff
@@ -148,10 +151,11 @@
     "patch-package": "^6.4.7",
     "postinstall-postinstall": "^2.1.0",
     "prettier": "^2.4.1",
+    "prop-types": "^15",
```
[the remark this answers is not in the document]
peoplespete: Ripping it out. 
I added because of this warning from yarn:
```
warning " > react-idle-timer@4.6.4" has unmet peer dependency "prop-types@>=15".
```
But I've run cypress and jest tests without issue so I think good call to not have it.

services/ui-src/src/components/Inputs/RadioButton/index.tsx
```diff
@@ -76,7 +76,7 @@ export const RadioButton = ({
                 onDelete={option.onDelete}
               >
                 <CUI.Radio
-                  value={option.value}
+                  value={option.value as string}
```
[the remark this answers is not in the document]
peoplespete: 👍  Fixed.

services/ui-src/src/measures/2021/SSHH/index.test.tsx
peoplespete: Yeah I read this. I am looking at whether we really need all these changes at all. For now I've backed them out (including the jest retry 3 times thing). But there does seem to be a new flakeyness with the tests.

I backed out all of the changes and now the tests break (though they work mostly if you add back the jest retry 3 times). 
All of them are measures tests like this one for example:
`src/measures/2022/CCPAD/index.test.tsx`

The problem is that sometimes it won't find the test id element:
```
  it("measure should render", async () => {
    useApiMock(apiData);
    renderWithHookForm(component);
    expect(screen.queryByTestId("measure-wrapper-form")).toBeInTheDocument(); // <<< this one!
    await waitFor(() => {
      expect(screen.getByText(measureAbbr + " - " + description));
    });
  });
```

I've tried changing it to findByTestId and that didn't help for some reason. (I kind of thought that it would since I figured that was pretty much equivalent to what I did before by moving the queryBy into the waitFor but I might just be wrong about how similar those are).

Honestly I don't really understand enough about the internals of react rendering and RTL to actually understand why the updated packages are causing this issue. If you can think of a good idea I'm all ears. Short of that, I'm thinking of adding back the jest retry and putting in a tech debt ticket to fix it later. What do you think?

services/ui-src/setupJest.ts
```diff
@@ -3,6 +3,20 @@ import "@testing-library/jest-dom";
...
+    media: query,
+    onchange: null,
+    addListener: jest.fn(), // Deprecated
+    removeListener: jest.fn(), // Deprecated
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn(),
+  })),
+});
+jest.retryTimes(3);
```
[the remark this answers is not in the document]
peoplespete: It's a ton of the measure ones.
I wrote up a more longer winded explanation in response to the comments you made about RTL best practices.

services/ui-src/src/measures/2023/NCIDDSAD/index.test.tsx
```diff
@@ -74,7 +74,7 @@ describe(`Test FFY ${year} ${measureAbbr}`, () => {
     renderWithHookForm(component);
     expect(screen.getByTestId("measure-wrapper-form")).toBeInTheDocument();
     await waitFor(() => {
-      expect(screen.getByText(measureAbbr + " - " + description));
+      expect(screen.getAllByText(measureAbbr + " - " + description));
```
peoplespete, line 77: This seems to need to be there because of some chakra thing where we have links and chakra test both that have this text on them:
image

services/ui-src/src/views/AddStateSpecificMeasure/index.test.tsx
```diff
@@ -61,14 +61,18 @@ describe("AddStateSpecificMeasure", () => {
         screen.getByText(/\+ Add Another/i).click();
       }
 
-      // Expect that we can add a 5th SSM
-      expect(screen.getAllByText(/Name the measure/i)).toHaveLength(4);
-      expect(screen.getByText(/\+ Add Another/i)).toBeEnabled();
+      waitFor(() => {
```
peoplespete, line 64: @benmartin-coforma This is another perplexing RTL case. This time I figured out I need to add an await to findAllByText. But even when I did that the test wouldn't pass unless I do this waitFor wrapper. Does that make any sense to you?

services/ui-src/src/components/Banner/FormElements/TextField.tsx
```diff
@@ -34,7 +34,7 @@ export const TextField = ({
       label={label || ""}
       placeholder={placeholder}
       onChange={(e) => onChangeHandler(e)}
-      errorMessage={errorMessage}
+      errorMessage={errorMessage as React.ReactNode}
```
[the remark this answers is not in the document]
peoplespete: Looks like this is another place where toString will fix it. Will push up soon.

services/ui-src/src/components/Layouts/StateLayout/index.tsx
```diff
@@ -6,6 +6,7 @@ import { BreadCrumbItems } from "components";
 interface Props {
   breadcrumbItems: BreadCrumbItems;
   buttons?: ReactElement<any, any>;
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
peoplespete: Found a more elegant way to do this. It's one of the react 18 changes I believe:
https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc

Will push up shortly.

services/ui-src/src/components/Banner/FormElements/TextField.tsx
```diff
@@ -46,5 +46,6 @@ interface Props {
   placeholder?: string;
   nested?: boolean;
   [key: string]: any;
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
peoplespete: Changing to be slightly more elegant. Related to React 18 version upgrade.

services/ui-src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <link rel="icon" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="stylesheet" src="/css/vendor/design-system/index.css" />
+    <link rel="stylesheet" href="https://design.cms.gov/cdn/design-system/7.0.0/css/core-theme.css" />
```
benmartin-coforma, line 7: This is probably fine, but it may impact our PDF rendering (which actually happens on a 3rd party's server). @angelaco11 , do you remember how to test the print view? I can't remember if there's just one specific page that uses DocRaptor's rendering, or if all of the print buttons on the site do.

services/ui-src/src/measures/2021/SSHH/index.test.tsx
benmartin-coforma: Yeah, I'm fine kicking this one down the road. I'm only somewhat familiar with the internals myself, and I do not have the time to do a more thorough investigation.

services/ui-src/src/measures/2023/NCIDDSAD/index.test.tsx
benmartin-coforma: That's not too surprising, and this is a good way to handle it.

benmartin-coforma (approved):
Thanks for going through all of this, Pete!

services/ui-src/src/views/AddStateSpecificMeasure/index.test.tsx
benmartin-coforma: Could be a kind of """frame lag""" type thing? Since we're clicking the button three times, it takes three render cycles to display everything we expect? Aaand maybe findAllByText was just executing immediately, even when awaited? I'm not too fussed about it here as a one-off though.

services/ui-src/index.html
peoplespete: This is from this other PR (https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2512/files)
Just went through this with Aileen and she said the export and print both resulted in good looking pages.

ailZhou (commented):
_found a bug while I was fixing the css issue. when you save a change in the measure, there is a little note that tells you when you have last saved that particular measure. For some reason, that message is now showing it for each measure irregardless of whether it was saved there or not._ 

Ignore this, user error (the user is me)

ailZhou (commented):
also found this one when doing coreset add/delete (only 2023 and earlier)

https://github.com/user-attachments/assets/9703b81b-6cde-4963-ac71-df163fa7ae25

peoplespete:
> also found this one when doing coreset add/delete (only 2023 and earlier)
> 
>  Screen.Recording.2024-12-12.at.2.54.21.PM.mov

We figured out it is not related to the upgrade so Aileen has graciously agreed to write it up as a separate bug ticket.

ailZhou (approved):
lgtm! all we can hope is that there isn't any really bad bugs lying in hiding

benmartin-coforma (approved):

PR #2515, opened 2024-12-10
(2021) Remove Index.tsx from Measures

ailZhou:
### Description

This is the PR that will finally kickstart our move:
Screenshot 2025-01-10 at 11 57 39 AM

About 13 measures could not be moved yet. These measures that will be updated in a separate PR:
- **AIF-HH, IU-HH** - Changing these measures requires changing the OMS section. I will have a PR that will handle these for all 5 years.
- **LBW-CH, LRCD-CH, NCIDDS-AD, PDENT-CH** - These are all autocomplete measures. There currently is one in QMR but it will require some refactoring to match what we are doing here, so I'm going to save it for another PR, as it affects all the years and not just the current one.
- **AMR-AD, CPA-AD, CPC-CH, MSC-AD, PCR-AD, PCR-HH, SS-HH** - All have unique files that override the base components being used, so they will require a little bit more thought and possibly each have their own template file.
---
### So what is the secret sauce?
`measureTemplate.tsx` and `measureTemplateData.tsx`.

Majority of the measure do share the same page setup but the prop values they pass along are different which meant that that structure was the same but the data is different. Each measure already has a data.ts file, so all we had to do was expand it to hold more unique data than before.

example: AMB-CH moving data from _AMBCH/index.tsx_ to _AMBCH/data.ts_ 
Screenshot 2025-01-10 at 12 10 19 PM

Doing this allows up to reduce the _index.tsx_ files into a generic _measureTemplate.tsx_ that handles displaying the data but holds no data.

**Main files with changes**
_measureTemplate.tsx_ - Really the only file worth looking at.
_measureTemplateData.tsx_ - what data we are loading to the measure template.
_shared/types/MeasureTemplate.tsx_ - the types of the new measure data. This is mostly for Ben
_measures/2021/index.tsx_ - How we link the measureTemplate to the app.

All the other files are stripping the relevant information from the index.tsx and moving them to the data.ts and minor adjustment to that validation's object call

### Related ticket(s)

CMDCT-4177

---
### How to test

Deploy Link: https://d18g5nn2688622.cloudfront.net/

1) Sign into QMR, any state user.
2) Select reporting year 2021.
3) Create a combined Child Core Set.
4) We will check two measures to show that the template is working. ADD-CH & AMD-CH. Both are using the new template but will be slightly different:

Data Source: ADD-CH 
Screenshot 2025-01-10 at 12 46 34 PM

Data Source: AMB-CH
Screenshot 2025-01-10 at 12 45 42 PM

Rate Calc: ADD-CH
Screenshot 2025-01-10 at 12 47 54 PM

Rate Calc: AMB-CH
Screenshot 2025-01-10 at 12 48 08 PM

5) Ignore filling out the form and validate to make sure the validation is working.
6) Fill out the form and then validate. Refresh the page to make sure the data is saving.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


services/ui-src/src/measures/2021/CCPCH/data.ts
```diff
@@ -1,16 +1,24 @@
-import { DataDrivenTypes } from "shared/types";
+// import { DataDrivenTypes } from "shared/types";
```
rocio-desantiago, line 1: Do we still need this comment?

BearHanded:
My only question on this is after the change, if a measure turns out to need something entirely new and radical, what is the path for setting that up?

BearHanded (approved):

PR #2560, opened 2025-02-10
PQI Combined Rates fix

benmartin-coforma:
### Description
This PR addresses an issue introduced by PR #2386.

Context: PQI rates differ from most. Rather than a percentage, the rate is displayed as per hundred thousand. Prior to the refactor in 2386, our combined rate code would use the 100k multiplier if the first three letters of the measure abbreviation were PQI. After my refactor, we only use that multiplier if the measure abbreviation is `PQI-AD` exactly. Unfortunately, there is no such measure. There are actually _four_ PQI measures: `PQI01-AD`, `PQI05-AD`, `PQI08-AD`, and `PQI15-AD`. This is now reflected in the code.

Since that refactor was done after UAT testing had begun, but before the feature went live, all existing combined rates for PQI measures in production are incorrect. Thus, this PR also contains a migration to recalculate all of those values. This can be done pretty neatly, because the calculation is entirely server-side: Read the separate values from the measures table, perform the calculation, and write the result to the rates table.

### Related ticket(s)
CMDCT-

---
### How to test
1. Log in to QMR as a state user, for a state that uses combined rates
2. Enter some values for PQI01-AD in the CHIP core set. Note the calculated rate value.
3. Enter some values for PQI01-AD in the Medicaid core set. Note the calculated rate value.
4. Go to the combined rates page for PQI01-AD. The calculated combined rate should be about the same as the individual values. Notably, it should _not_ be 1000 times smaller.
5. I guess poke around some of the other PQI measures? If you like.

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/handlers/rate/migratePqiMiscalculations.ts
+++ b/services/app-api/handlers/rate/migratePqiMiscalculations.ts
@@ -27,8 +27,6 @@ export const main = async () => {
     );
     logger.info(`Found ${allExistingRates.length} existing rates`);
 
-    // The new shape for combined rates has an "AdditionalValues" property.
-    // If a rate doesn't have that property, it's the old shape.
     const ratesNeedingMigration = allExistingRates.filter((rate) =>
       rate.measure.startsWith("PQI")
     );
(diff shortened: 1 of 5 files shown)


rocio-desantiago (dismissed):
Tested `2025 PQI01-AD Combined Rates` if anyone wants to test the others 😃 
Combined Rates displayed as follows:

BearHanded (approved):

rocio-desantiago (approved):

BearHanded (approved):

ailZhou (approved):
code looks good, and tested in deploy measure PQI05-AD

PR #2584, opened 2025-03-28
CMDCT-4420: fixing pdf display regressions after chakra updates

angelaco11:
### Description
A long time ago, I did some horrible stuff to brute force some css for pdf export since prince pdf has all sorts of issues and limitations with chakra elements. I did this by inspecting the checkbox and radio elements in the html pdf export page, grabbing the css class names (usually in the format of css-1jbejj2 or whatever) and then setting css manually in the util.tsx that you see in this PR. Well, chakra updates caused chakra css classes to change. Since the pdf export needs to override the chakra css and is grabbing the specific checkbox and radio css class names, I am updating the changed chakra classes to reflect the new classes.

This PR and my initial work to fix PDF ([here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/2316)) should be referenced any time qmr pdfs have a regression in display. I also tried to leave helpful and descriptive comments. 

### Related ticket(s)
CMDCT-4420

---
### How to test
https://d2oahmgqgfu93v.cloudfront.net/
1. Go to QMR core set view and click on the three dot menu of any core sets where you've filled out some measures (i would recommend filling out the first few so you don't have to scroll to far in pdf) and click "Export"
2. This will take you to the pdf print out page. Click the "Print PDF" button and wait for the PDF to export
3. Once the PDF opens in a new tab, give it a good look-see. Make sure it sort of looks as good as possible and make sure that checkboxes and radio buttons look correct for the active ones

### DIFF
--- a/services/ui-src/src/views/ExportAll/util.tsx
+++ b/services/ui-src/src/views/ExportAll/util.tsx
@@ -127,7 +127,9 @@ export const applyPrinceSpecificCss = (): HTMLStyleElement => {
     
     ${
       /* ******* Prince doesn't support certain css elements like background colors or css grid out of the box. we have to brute force those styles here.
-      All of these css class names can be found by inspecting the elements on the pdf page in the browser. ********* */ ""
+      IMPORTANT NOTE: If there is a display regression, it's probably because these classes changed due to chakra updates or other unknowns.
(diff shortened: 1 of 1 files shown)


ailZhou (approved):
Angela!!! YOU'RE THE BEST, NEVER LEAVE!

rocio-desantiago (approved):
Amazing

PR #2590, opened 2025-04-24
CMDCT-4363 converting to cdk

peoplespete:
### Description
Converts QMR to a working CDK app

### Related ticket(s)
CMDCT-4363

---
### How to test
You should be able to use the application here:
https://d1cw8jmupnbtnz.cloudfront.net/
As long as you're Zscaling :)

### Notes
NA, I've called out various things in the code with comments.

---
### Pre-review checklist

---
#### Review

#### Security

---


deployment/prerequisites.ts
```diff
@@ -0,0 +1,91 @@
...
+      {
+        assumedBy: new iam.ServicePrincipal("apigateway.amazonaws.com"),
+        permissionsBoundary: isLocalStack
+          ? undefined
+          : iam.ManagedPolicy.fromManagedPolicyArn(
+              this,
+              "iamPermissionsBoundary",
+              `arn:aws:iam::${Aws.ACCOUNT_ID}:policy/cms-cloud-admin/developer-boundary-policy`
+            ),
+        path: "/delegatedadmin/developer/",
```
peoplespete, line 46: @JonHolman Do you have an opinion on whether I should go ahead and rip all these out now or wait and do it in a follow up ticket?

JonHolman: @peoplespete since it's not merged yet, I'd rip it out now.

deployment/stacks/uploads.ts
```diff
@@ -0,0 +1,319 @@
...
+          s3.HttpMethods.DELETE,
+          s3.HttpMethods.HEAD,
+        ],
+        allowedHeaders: ["*"],
+        exposedHeaders: ["ETag"],
+        maxAge: 3000, // 50 minutes
+      },
+    ],
+  });
+
```
peoplespete, line 120: If you look _really really_ close, you'll notice that before we also granted some permissions to the shared lambda execution role. I could find no evidence that they were ever needed or used so I removed. Also our cdk setup gives each lambda it's own execution role instead of sharing.

deployment/stacks/uploads.ts
```diff
@@ -0,0 +1,319 @@
...
+    blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+    objectOwnership: s3.ObjectOwnership.BUCKET_OWNER_PREFERRED,
+    removalPolicy: isDev ? RemovalPolicy.DESTROY : RemovalPolicy.RETAIN,
+    versioned: true,
+    enforceSSL: true,
+  });
+
+  const clamDefsBucket = new s3.Bucket(scope, "ClamDefsBucket", {
+    bucketName: `${service}-${stage}-avscan-${Aws.ACCOUNT_ID}`,
+    encryption: s3.BucketEncryption.S3_MANAGED,
```
peoplespete, line 180: Note that a couple buckets use a KMS key for encryption and some let s3 manage this. I assume that's because the folks who injest data need access to things like dynamo snapshots and attachments but not the other things we put in buckets.

services/app-api/handlers/coreSets/get.ts
```diff
@@ -14,7 +14,7 @@ import {
   parseCoreSetParameters,
 } from "../../utils/parseParameters";
 
-export const coreSetList = handler(async (event, context) => {
+export const getAllCoreSets = handler(async (event, context) => {
```
peoplespete, line 17: Why change the name of this? There was a pattern with names of things in the definition of the api and this was the only one that was not following the pattern.

deployment/stacks/ui.ts
```diff
@@ -0,0 +1,249 @@
...
+) {
+  const githubIpSet = new wafv2.CfnIPSet(scope, "GitHubIPSet", {
+    name: `${stage}-gh-ipset`,
+    scope: "CLOUDFRONT",
+    addresses: [],
+    ipAddressVersion: "IPV4",
+  });
+
+  const wafRules: wafv2.CfnWebACL.RuleProperty[] = [];
+  if (isDev && vpnIpSetArn && vpnIpv6SetArn) {
```
peoplespete, line 190: isDev will always mean that vpnIpSetArn and vpnIpv6SetArn are there but I included them in the if for good measure in case later someone changes the config accidentally.

peoplespete: Also @BearHanded isDev feels like the right if here (ephemeral apps will be behind vpn, main/val/prod officially wired up urls will not). In the past it was done on an AWS account basis, which I think was probably simpler with the sls setup.

deployment/prerequisites.ts
peoplespete: Just did it! Most exciting moment was realizing we didn't need addIamPropertiesToBucketRole at all. Thanks for paving the way on this one.

deployment/stacks/ui-auth.ts
```diff
@@ -0,0 +1,323 @@
+import { Construct } from "constructs";
+import {
+  aws_cognito as cognito,
+  aws_iam as iam,
+  aws_lambda as lambda,
+  aws_lambda_nodejs as lambda_nodejs,
+  aws_wafv2 as wafv2,
+  aws_ssm as ssm,
```
JonHolman, line 8: SSM?

src/run.ts
JonHolman, line 1: local run needs to invoke the seed data function, right?

services/uploads/src/dynamoSync/handler.js
```diff
@@ -54,7 +54,7 @@ const cleanupFolders = async () => {
 
     if (outdated.length <= 0) continue;
     const deleteParams = {
-      Bucket: process.env.dynamoSnapshotS3BucketName,
+      Bucket: process.env.DYNAMO_BUCKET_NAME,
```
JonHolman, line 57: Do you want to match the casing of what we're now doing for the dynamo tables for consistency, so `DynamoBucket` which now I say to myself doesn't make sense :) I mean NameBucket would make sense for consistency but having a bucket named Dynamo I don't understand.

deployment/stacks/imports_included/data.ts
```diff
@@ -0,0 +1,54 @@
+import { Construct } from "constructs";
+import { aws_dynamodb as dynamodb, aws_s3 as s3, Tags } from "aws-cdk-lib";
```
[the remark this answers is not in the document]
peoplespete: Fixing 👍

deployment/stacks/ui-auth.ts
peoplespete: blurp...
fixing!

services/uploads/src/dynamoSync/handler.js
peoplespete: I'd like to leave it this way unless we overhaul all the env variables. There are still a bunch of env variables cased this way in .env and .env.tpl so for now I'd like to have 2 ways of doing it and the only way you do not THIS_WAY is if you are the name of a table. But I'm open to making them all ThisWay down the road. You ok with that?

deployment/stacks/ui.ts
```diff
@@ -0,0 +1,253 @@
...
+    enforceSSL: true,
+    blockPublicAccess: s3.BlockPublicAccess.BLOCK_ALL,
+    versioned: true,
+  });
+
+  let loggingConfig:
+    | { enableLogging: boolean; logBucket: s3.Bucket; logFilePrefix: string }
+    | undefined;
+  if (!isDev) {
+    const logBucket = new s3.Bucket(scope, "CloudfrontLogBucket", {
```
peoplespete, line 53: @BearHanded @benmartin-coforma @JonHolman 
Finally made some progress on why stack deletions were occasionally failing. First I needed to add autoDeleteObjects to a couple buckets that were missing it. That still left ~66% of the time failures and always because of cloudfront logging bucket. I figured out that it was because the autoDeleteObjects doesn't do a great job when things are still attached and logging (potentially esp when the bucket in question is versioned). The code for the custom resource that cdk creates to manage this has some code that tries to block continued puts on the bucket but comments surrounding the code did not inspire confidence.

Soooo what if for ephemeral environments we don't do cloudfront logging? I've tested it 6 times and it's worked every time!

JonHolman: Sounds good to me.

JonHolman (approved):

BearHanded (approved):

src/run.ts
peoplespete: Brendan confirmed that QMR doesn't need seeding at this point.

PR #2624, opened 2025-05-08
[QMR] Remove Public Reporting Hint Text (CPA-AD & CPC-CH)

rocio-desantiago:
### Related ticket(s)

CMDCT-4495

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/measures/2025/CPAAD/index.tsx
+++ b/services/ui-src/src/measures/2025/CPAAD/index.tsx
@@ -1,88 +1,10 @@
 import * as Q from "./questions";
-import * as CMQ from "shared/commonQuestions";
 import { useParams } from "react-router-dom";
 import * as QMR from "components";
-import * as CUI from "@chakra-ui/react";
 import { useFormContext } from "react-hook-form";
 import { FormData } from "./types";
 import { validationFunctions } from "./validation";
 import { useEffect } from "react";
-
-const defOfDenomOptions: CMQ.CoreSetSpecificOptions = {
-  ACSM: {
-    options: [
-      {
-        displayValue: "Medicaid (Title XIX)",
-        value: "Medicaid (Title XIX)",
-      },
-      {
-        displayValue: "Medicaid-Expansion CHIP (Title XXI)",
-        value: "SurveySampleIncCHIP",
-      },
-      {
-        displayValue: "Individuals Dually Eligible for Medicare and Medicaid",
-        value: "SurveySampleIncMedicareMedicaidDualEligible",
-      },
-    ],
-    helpText: (
-      <>
-        <CUI.Text mt="3" mb="3">
-          Please select all populations that are included in the survey sample.
-          For example, if your survey sample includes both non-dual Medicaid
-          (Title XIX) beneficiaries and Individuals Dually Eligible for Medicare
-          and Medicaid, select both Medicaid population (Title XIX) and
-          Individuals Dually Eligible for Medicare and Medicaid.
-        </CUI.Text>
-      </>
-    ),
-  },
-  ACS: {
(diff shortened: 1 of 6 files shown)


benmartin-coforma (dismissed):
I haven't tested, but this makes sense to me!

services/ui-src/src/measures/2025/CPAAD/index.tsx
```diff
@@ -110,14 +32,11 @@ export const CPAAD = ({
         <>
           <Q.HowDidYouReport />
           <Q.DataSource type="adult" />
-          <CMQ.DefinitionOfPopulation
-            coreSetOptions={defOfDenomOptions}
-            deliverySystems={false}
-          />
+          <Q.DefinitionOfPopulation deliverySystems={false} />
```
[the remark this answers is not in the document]
ailZhou: If you want to fully change to using the custom DefinitionOfPopulation file, double check to make sure it isn't missing any functionality that it was using in the shared component. The annoying part is that you have to check for uniqueness in both Medicaid and CHIP versions.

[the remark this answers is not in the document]
rocio-desantiago: Updated. 😄

services/ui-src/src/measures/2025/CPAAD/questions/DataSource.tsx
```diff
@@ -28,16 +28,7 @@ export const DataSource = ({ type }: DataSourceProps) => {
...
-                  <>
-                    Describe the data source (
-                    <em>
-                      text in this field is included in publicly-reported
-                      state-specific comments
-                    </em>
-                    ):
-                  </>
-                }
+                label={<>Describe the data source:</>}
```
benmartin-coforma, line 31: Tiny suggestion: We can get away with just a string here.

```suggestion
                label="Describe the data source:"
```

services/ui-src/src/shared/commonQuestions/DefinitionsOfPopulation.tsx
```diff
@@ -305,28 +307,30 @@ const CoreSetSpecificDefinitions = (
...
+      value: DC.DENOMINATOR_INC_OTHER,
+      children: [
+        <QMR.TextArea
+          formLabelProps={{ fontWeight: "400" }}
+          label={parseLabelToHTML(labels.defineDenomOther)}
+          {...register(DC.DEFINITION_DENOMINATOR_OTHER)}
+        />,
+      ],
+    });
+  }
```
benmartin-coforma, line 327: Thought: I usually like to initialize variables directly into their final state, whenever it's possible & reasonable. In this case, it's not too much trouble to initialize the `otherOptions` as having either 0 or 1 elements - versus initializing with 0 and maybe adding 1.

Writing it this way has the added benefit of removing the negation from the condition - `if`s and other conditionals are always a little easier for me to read when they don't include a `!`.

How do you like this?

```suggestion
  const otherOption = {
    displayValue: "Other",
    value: DC.DENOMINATOR_INC_OTHER,
    children: [
      <QMR.TextArea
        formLabelProps={{ fontWeight: "400" }}
        label={parseLabelToHTML(labels.defineDenomOther)}
        {...register(DC.DEFINITION_DENOMINATOR_OTHER)}
      />,
    ],
  };
  const otherOptions = removeOtherOption ? [] : [otherOption];
```

rocio-desantiago: Ooh I knew there existed a cleaner way to write that 😄

ailZhou (approved):

PR #2636, opened 2025-05-19
Add some unit tests around ExportAll

benmartin-coforma:
### Description
It does what it says.

`getSpaName`, I refactored for clarity. The test for spaLib, I added so that I could justify excluding the check previously on line 222 in `getSpaName`. It was checking to see, not only whether it found an object, but also whether that object had truthy id, name, and state. If those are always truthy for every object in `SPA`, then that check is unneeded. And the unit test for spaLib asserts that truthiness.

I would have liked to add more tests around the helper functions within util.tsx. But I wanted to try to test them from the actual call site, if possible. Helpers like `openPdf` and `cloneChakraVariables` are only ever called from within `usePrinceRequest`, so I wanted to test them by calling them indirectly. But writing tests for `usePrinceRequest` is hard, because it calls React Hooks, and you're not allowed to call hooks unless you're already in a hook - which a unit test is not.

It is possible to work around this - for example, by using `react-hooks-testing-library`. But [they discourage](https://github.com/testing-library/react-hooks-testing-library?tab=readme-ov-file#when-to-use-this-library) use of their own tool when avoidable. And in theory, it is avoidable here - by writing tests for the ExportAll page, and clicking the button to call the hook.

But by the time I'd written some tests for the page and was ready to dig in to test the Prince hook, I had sunk in enough time to make this PR and call it a day. I'll get you next time, `usePrinceRequest` 💪 

### Related ticket(s)
n/a

---
### How to test

https://dtq45arh0fflf.cloudfront.net/

The unit tests should pass.

Visiting the ExportAll page for a Health Home core set should display the correct SPA name.

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/libs/spaLib.test.ts
+++ b/services/ui-src/src/libs/spaLib.test.ts
@@ -1,0 +1,17 @@
+import { SPA } from "./spaLib";
+
+describe("spaLib", () => {
+  it("Should contain properly-shaped data", () => {
+    // Technically, this test is redundant with the Typescript definitions.
+    for (let yearlySpas of Object.values(SPA)) {
+      for (let spa of yearlySpas) {
+        expect(typeof spa.id).toBe("string");
+        expect(typeof spa.state).toBe("string");
+        expect(typeof spa.name).toBe("string");
+        expect(spa.id).toBeTruthy();
+        expect(spa.state).toBeTruthy();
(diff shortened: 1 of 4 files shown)


ailZhou (approved):
tested the spa refactor in deploy and it looks good.

PR #2704, opened 2025-08-01
[CMDCT-4872] Maintain Expanded/Collapsed State on Save/Validate

ailZhou:
### Description

UPDATE 8/6/25: Bandaid not needed! @BearHanded helped me pinpoint which line was the closest to the problem. There is still a bigger problem of rebuilding data constantly but that's not important for this ticket. The solution is a lot more straightforward, two useStates. I am keeping the `getFilledKeys` function as I feel like this maybe useful in the future but if not, I'll remove it at the end of 2026 year.

Accidentally deleted the previous summary, this PR is to fix the collapsing issue that happens with the user triggers a save in the data. There is still a rerendering sometimes but the accordions should be in the same state as where you left it.

### Related ticket(s)

CMDCT-4872

---
### How to test

Deploy Link: https://d3e99w0gavycca.cloudfront.net/

1) Sign into QMR, any state user (i.e. [stateuser1@test.com](mailto:stateuser1@test.com))
2) In reporting year 2025, select any measure that has a **Measure Stratification** section (i.e. WCC-CH)
3) In the **Measure Specification** section, select the first radio button, this will enable the Performance Measure section
4) Scroll down to the **Performance Measure** section and fill out the N/D/R sets. This will enable the **Measure Stratification** section
5) Go to the  **Measure Stratification** section and in the radio button section select 2024
6) Fill out an N/D/R set with different denominators. 
7) Validate and make sure that the accordion stays the same when you save.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/Accordion/index.tsx
+++ b/services/ui-src/src/components/Accordion/index.tsx
@@ -47,15 +47,11 @@ export const Accordion = ({
   const [index, setIndex] = useState<number>(1);
   const shared: any = useContext(SharedContext);
 
+  //this useEffect is activated when a measure is saved. it is used to expand the accordion to show stratification that is filled
   useEffect(() => {
-    if (externalControlled && shared?.stratification?.length > 0) {
-      if (id) {
-        const keys = shared?.stratification;
-        for (var i = 0; i < keys.length; i++) {
-          if (keys[i].includes(id)) {
-            setIndex(0);
-          }
-        }
+    if (id && shared?.stratification?.length > 0) {
+      if (shared?.stratification.some((key: string) => key.includes(id))) {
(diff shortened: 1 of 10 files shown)


Filtos (approved):
Looks good! Small nit-pick: There is a weird jerk with the screen jumping to dropdown -> validation errors -> dropdown when you click "Validate".
Might be a good idea to check in with Cam to make sure its fine.

rocio-desantiago (approved):
Tested. Saw the same thing @Filtos noted about the little jump. Other than that it's good 👍

rocio-desantiago (approved):
No jump this time 😄

PR #2720, opened 2025-08-15
[CMDCT-5053]  ECDS "Not Answered" Bug

ailZhou:
### Description

This PR is to address the "Not Answered" text being shown in the Data Source display for combined rates. This only affects measures with Electronic Clinical Data Systems (ECDS) as that data source option has both a checkboxes and a textbox on the same group layer.
 
Screenshot 2025-08-18 at 12 37 31 PM

I rewrote the code to be a bit more readable compare to before. The dataSourceSelections object is converted into an array so that we can utilize filters and finds. It will only show not answered if there's nothing selected in the selections and nothing is filled out for the description

Note:
- The changed in index.html is a removal of a bootstrap call that is never used. Cypress had discovered it awhile but I forgot to remove it until now.
  
### Related ticket(s)

CMDCT-5053

---
### How to test

Deploy Link: https://d2dbhsolx986rn.cloudfront.net/

1) Sign into QMR, any state user
2) Go to any of these measure: AIS-AD
3) In the **Data Source** section, select all checkboxes and then fill out the textbox. 
4) Save
5) Go back and clear the textbox but leave the checkboxes so it should look like this now:
Screenshot 2025-08-18 at 12 21 37 PM
4) Save or validate
5) Go to the combined rates page of that measure
6) See that the Not Answered is not listed in the data source
Screenshot 2025-08-18 at 12 23 23 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/shared/commonQuestions/DataSouceInformationBanner/DataSourceInformationBanner.tsx
+++ b/services/ui-src/src/shared/commonQuestions/DataSouceInformationBanner/DataSourceInformationBanner.tsx
@@ -123,16 +123,20 @@ export const dataSourceSelections = (
 ) => {
   let selected = [];
 
-  const remapped = Object.entries(dataSourceSelections).map((selection) => {
-    return { key: selection[0], ...selection?.[1] };
-  });
+  //remapping the object to a more usable array
+  const dataSourceArray = Object.entries(dataSourceSelections).map(
+    (selection) => {
+      return { key: selection[0], ...selection?.[1] };
+    }
+  );
 
   //we want only they key for the top layer checkboxes and the current dataSource
-  const parentDataSource = remapped.filter(
+  const parentDataSource = dataSourceArray.filter(
     (selection) =>
       !selection.key.includes("-") && selection.key.includes(dataSource)
   );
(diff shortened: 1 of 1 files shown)


ailZhou:
> Not seeing this issue anymore for the first data source but if I follow those same steps for "Other Data Source" I am: Screenshot 2025-08-20 at 10 41 57 AM
> 
> not sure if the fix is supposed to also apply to the "other" option?

That's right for the Other option. The easiest way to think of this is if there's even one bullet of an answer there, than we don't show a Not Answered but if there's nothing there then we want to indicate that it's Not Answered.

ajaitasaini:
> > Not seeing this issue anymore for the first data source but if I follow those same steps for "Other Data Source" I am: Screenshot 2025-08-20 at 10 41 57 AM
> > not sure if the fix is supposed to also apply to the "other" option?
> 
> @ajaitasaini That's right for the Other option. The easiest way to think of this is if there's even one bullet of an answer there, than we don't show a Not Answered but if there's nothing there then we want to indicate that it's Not Answered.

gotcha, thank you for clarifying!!

ajaitasaini (approved):

PR #2723, opened 2025-08-18
2025 - Performance Measure Warning Text Change

ajaitasaini:
### Description

Change validation banner (in just the Adult separate CHIP report) to read: “Performance Measure Warning: Information has been included in the Age 65 and Older Performance Measure but the checkmark for (Denominator Includes Medicare and Separate CHIP Dually-Eligible population) is missing”

### Related ticket(s)

CMDCT-5051

---
### How to test

https://d59p56qgg1cx7.cloudfront.net/

1. Go to an Adult Separate Chip Measure
2. Go to CDF-AD or any other measure
3. Go down to the measure and fill out  N/D/R for Age 65 and Older
4. DON'T Click Individuals Dually Eligible for Medicare and Separate CHIP in Definition of Denominator section. 
5. Press validate, you should see this banner with the updated language
Screenshot 2025-08-19 at 2 46 15 PM

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

---


rocio-desantiago (dismissed):
tested with measure HVL-AD and warning pops up as expected

services/ui-src/src/shared/globalValidations/validateDualPopInformation/index.ts
```diff
@@ -1,8 +1,20 @@
...
 import { featuresByYear } from "utils/featuresByYear";
+import { CoreSetAbbr } from "types";
 
-export const getLabels = (errorReplacementText: string) => {
+export const getLabels = (errorReplacementText: string, coreSet?: string) => {
+  if (
+    coreSet &&
+    coreSet === CoreSetAbbr.ACSC &&
+    featuresByYear.hasAdultSeparateCHIPInclusiveWarning &&
+    featuresByYear.shouldValidateDuallyEligibleCheckbox
```
Filtos, line 11: Do we need this check here, if `featuresByYear.hasAdultSeparateCHIPInclusiveWarning` checks that the year is already >= 2025, so `featuresByYear.shouldValidateDuallyEligibleCheckbox` would always be true.

services/ui-src/src/shared/globalValidations/validateDualPopInformation/index.ts
```diff
@@ -19,13 +31,14 @@ export const validateDualPopInformationPM = (
   OPM: any,
   age65PlusIndex: number,
   DefinitionOfDenominator: string[] | undefined,
+  coreSet?: string,
```
Filtos, line 34: Tiny nit-pick but I think its better practice to put the optional params last, this way you wont have to pass all those empty strings.

Filtos (approved):
Verified that implementation works, just left a few comments.

services/ui-src/src/shared/globalValidations/validateDualPopInformation/index.ts
ajaitasaini: I think in this case, it made sense to me to have the coreset param be in this spot, because the parameter that comes below this is an error label with a default value. If the coreset param was last in the params list, I'd have to pass in undefined everywhere for the error label (when the default value was desired for the error label) when a coreset is also passed through, and personally it made more sense to me to switch the order of the params

ailZhou (approved):
lgtm! i tested each messages and the validation message is changing correctly between chip and medicaid.

BearHanded (approved):

PR #2724, opened 2025-08-18
[QMR] 2025 - Qualifier Question Table Name Validation Error 

rocio-desantiago:
### Description

This PR covers adding a more general error message for when columns don't total 100%.
Previous years will continue with the specific errors that contain the age group in the error message as such: 

(2024 and earlier)
> "Entries for **Ages 21 to 64 column** must have values"

 (2025)

> "Entries for column must total 100"

Additionally, some may have multiple columns and would render an error message per column up to 3, and this refactors that portion for 2025 so that we don't see the same error message 3 times for that particular case in Adult Medicaid and Health Home qualifier table.

We are using featuresByYear to target this message to only show for 2025 and later. For previous years, it is important that the error messages stay year specific because the column name functionality and things like that change year to year. For example, some states have different age ranges for child vs adult columns.

### Related ticket(s)

CMDCT-5054

---
### How to test

**Deploy Link:** https://d1ak3v0ngff926.cloudfront.net/

1. Sign in with different users: stateuser1@test.com (combined rates) & stateuserDC@test.com (health home core sets) & alaska@test.com (state without combined rates) 
   - or any other state without combined rates from this list = ["AK", "AS", "GU", "HI", "NH", "NM", "NC", "ND", "OH", "PR", "SC", "VI", "VT", "WY"] from user list [here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/main/services/ui-auth/libs/users.json)
2. For each core set, click into the Core Set Qualifiers Questions
     Click-on-measures
     
     Click-Qualifier-Link

3. Fill out the table to total 100 to validate and <100 to confirm errors
      Validate-100
     
      Validate-Less-Than-100

4. Repeat this for years prior to confirm update only applies to 2025 (Below image is from Health Home core set)
      Prior year error messages for health homes

---
### Pre-review checklist

#### Review


services/ui-src/src/shared/Qualifiers/validations/adultChip.ts
```diff
@@ -13,13 +14,6 @@ const validate21To64EqualsToOneHundredPercent = (data: ACSCQualifierForm) => {
     0
   );
 
-  const total64PlusPercent = values?.reduce(
```
rocio-desantiago, line 16: this age range doesn't exist for adultChip

ailZhou (commented):
It looks like if I have a Total of 99%, I'm still able to get a complete message? This is in Adult Core Set Qualifiers: Medicaid 

Screenshot 2025-08-26 at 9 36 46 AM

ailZhou (approved):
LGTM! validation error message is now very general when values don't add up to 100%. Tested for both combined, separate and health homes. 

There is a label issue not related to your PR that i will address once your PR is merged in.

ajaitasaini (approved):
tested and lgtm! the only thing that I am seeing that I wasn't sure about was that if there are 2 columns that have the issue, is the error supposed to visually pop up twice? The screenshot in the testing instructions made it seem like that was the case. 

I did confirm that the error stays until both columns are addressed, so marking this non-blocking

ajaitasaini:
> > tested and lgtm! the only thing that I am seeing that I wasn't sure about was that if there are 2 columns that have the issue, is the error supposed to visually pop up twice? The screenshot in the testing instructions made it seem like that was the case.
> > I did confirm that the error stays until both columns are addressed, so marking this non-blocking
> 
> Good question. I asked Cam about this to confirm if we wanted to do just one message instead of 2 or 3 in the case of those with more than one column. My bad for the poor explanation. I put that screenshot as a before but I could see how that may look like what's expected. 🤗

omg no worries! Thank you for clarifying @rocio-desantiago

PR #2725, opened 2025-08-18
Allow numerator > denominator in Combined Rates for certain measures

benmartin-coforma:
### Description
Every measure in QMR has a fantastic, comprehensive amount of validation on the frontend. However, even data that has not passed validation can still be used for the Combined Rates page. When calculating combined rates, we just use whatever data was saved, without performing any validation - with one exception. We _do_ check to make sure that the numerator is not greater than the denominator. This is a simple check that applies to all measures, so it was not a huge cost to re-implement it within the combined rates code.

Well now, even that exception has an exception. There are a handful of measures to which that check does not apply. For EDV and PQI measures, it is possible and valid for the numerator to exceed the denominator. Most of our measures are about subsets (like, "Of the eligible population, how many received the service?"). But EDV and PQI aren't subsets, they're more general ratios (like "How many incidents per month?").

This PR implements that exception. We will now calculate combined rates _even if_ the numerator exceeds the denominator, for EDV and PQI measures only.

### Related ticket(s)
CMDCT-5020

---
### How to test
https://d28fa8f4f8vlo.cloudfront.net/

1. Log in as a state user, for a state with combined rates. For example, `stateuser1@test.com`
2. Confirm that "normal" measures still work as expected - standard calculation
   1. Fill out a normal measure, for both Medicaid and CHIP. For example, `AAB-CH`. Use an administrative data source for both individual measures.
   2. Note that the combined rate is calculated
   3. Change one of the individual measures to have numerator > denominator. You will see a warning on the page, but the data will still be saved when you click the Save button.
   4. Note that the combined rate now excludes that improper fraction.
3. Confirm that "normal" measures still work as expected - weighted calculation
   1. Fill out a normal measure that allows for a non-admin data source, for both Medicaid and CHIP. For example, `CPU-AD` allows Case Management Record Review.
   2. Note that the combined rate is calculated
   3. Change one of the individual measures to have numerator > denominator. You will see a warning on the page, but the data will still be saved when you click the Save button.
   4. Note that the combined rate now excludes that improper fraction.
4. Confirm that special measures are now allowed to be improper
   1. Fill out `EDV-AD`, for both Medicaid and CHIP.
   2. Note that the combined rate is calculated.
   3. Increase one of the numerators to be greater than the denominator.
   4. Note that the data is still included in the combined rate calculation.
5. (Testing the weighted calculation for special measures is not possible from the UI; neither EDV nor PQI allows hybrid data sources)

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/handlers/rate/rateNDRCalculations.test.ts
+++ b/services/app-api/handlers/rate/rateNDRCalculations.test.ts
@@ -366,4 +366,273 @@ describe("NDR calculations for Combined Rates", () => {
       },
     ]);
   });
+
+  it("Should refuse to calculate if numerator > denominator (standard formula)", () => {
+    const dataSources = {
+      Medicaid: {},
+      CHIP: {},
+    } as CombinedRatesPayload["DataSources"];
+
+    const medicaidMeasure = {
+      data: {
+        PerformanceMeasure: {
+          rates: {
+            cat0: [
+              {
+                uid: "cat0.qual0",
+                label: "mock rate",
+                numerator: "12",
+                denominator: "10",
+                rate: "20",
+              },
+            ],
+          },
+        },
+      } as Measure["data"],
+    } as Measure;
+
+    const chipMeasure = {
+      data: {
+        PerformanceMeasure: {
+          rates: {
+            cat0: [
+              {
+                uid: "cat0.qual0",
+                label: "mock rate",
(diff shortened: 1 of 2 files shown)


BearHanded (approved):
logic seems logical

PR #2730, opened 2025-08-21
Ensure a Complete core set cannot contain an In-Progress measure

benmartin-coforma:
### Description
Immediately after the UI successfully saves a measure, it sends a second call to the API to update the core set status to Incomplete. Unfortunately, "immediately after" is not soon enough. If the user navigates away before the measure update completes, the core set status update will never be sent. This can lead to a "Complete" or "Submitted" core set containing an "In Progress" measure, which is incorrect.

This PR fixes that situation, mostly. We now update the measure and the core set status in the same API call. It's still technically possible for one to succeed without the other - for example, if an exception is thrown from within `calculateAndPutRate()`. But that is far less likely than a user clicking around the site too fast. We can't entirely eliminate the possibility of this bug without switching to a transactional database; since we're staying on DynamoDB, this fix will have to be good enough.

There are a few other places in the app that can call the measure update endpoint; none of these will be adversely affected by the new code.

### Related ticket(s)
CMDCT-5065

---
### How to test
https://d28xmkm72evvpu.cloudfront.net/

The bug fixed by this PR is somewhat difficult to reproduce:
1. Complete a core set
2. Navigate to a measure page
3. Open your browser's dev tools, and enable throttling at a very low speed
4. Click the save button
5. Close your entire browser tab
6. Disable throttling, open a new tab, and go back to QMR. You should see that both the measure and core set status were updated (or, if you closed the tab _too_ quickly, neither).

### Notes
n/a

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/handlers/measures/tests/update.test.ts
+++ b/services/app-api/handlers/measures/tests/update.test.ts
@@ -8,6 +8,7 @@ jest.mock("../../../libs/dynamodb-lib", () => ({
   __esModule: true,
   default: {
     update: jest.fn(),
+    get: jest.fn(),
   },
 }));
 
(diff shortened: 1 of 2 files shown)


ailZhou (approved):
lgtm! i haven't been able to trigger the submitted issue so it looks to be fixed. thank you for fixing this bug!

rocio-desantiago:
Re-ran end to end tests failing.

rocio-desantiago (approved):
Cleared data. Refilled data. Saw it go back to in-progress and core set too!

PR #2735, opened 2025-08-26
[QMR] Missing header and footer landmarks

rocio-desantiago:
### Description

This PR adds header and footer with semantic HTML for Landmarks to be used in Assistive Tech (AT) for faster navigation.

### Related ticket(s)

CMDCT-4804

---
### How to test

**Deploy Link:** 
I checked the console, but I need to find out how to use VoiceOver to confirm this @cy-by

### DIFF
--- a/services/ui-src/src/components/Footer/index.tsx
+++ b/services/ui-src/src/components/Footer/index.tsx
@@ -98,46 +98,59 @@ export const emailLink = () => {
 export function Footer(): JSX.Element {
   return (
     <CUI.Box
+      as="footer"
+      role="contentinfo"
       zIndex={2}
       fontSize={{ base: "sm", md: "md" }}
       textAlign={{ md: "left" }}
       data-testid="footer"
       className="prince-measure-wrapper-box"
     >
-      <CUI.Container
-        data-testid="help-section"
-        className="help-section"
-        maxW="7xl"
-      >
-        <CUI.Box maxW="4xl">
-          <CUI.Text as="h3" fontSize="2xl" mb="3" className="prince-supp-text">
-            Do you have questions or need support?
-          </CUI.Text>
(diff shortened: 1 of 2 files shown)


benmartin-coforma (dismissed):
Code looks good to me!

Filtos:
I noticed that there is a header within the header now? From what Im reading that could be a problem, but we could check in with Cypress.
image

services/ui-src/src/components/Footer/index.tsx
```diff
@@ -98,46 +98,59 @@ export const emailLink = () => {
 export function Footer(): JSX.Element {
   return (
     <CUI.Box
+      as="footer"
+      role="contentinfo"
```
cy-by, line 102: The `role` doesn't need to be manually applied. The `<footer>` has this [role implicitly](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/contentinfo.html#:~:text=the%20html%20footer%20element%20defines%20a%20contentinfo%20landmark%20when%20its%20context%20is%20the%20body%20element).

cy-by:
> I noticed that there is a header within the header now? From what Im reading that could be a problem, but we could check in with Cypress. image
---
@rocio-desantiago @Filtos 
It is OK for the `<header>` to be inside of a `<header>`. It doesn't create two `banner` landmarks, the inner `<header>` [becomes a generic element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/header#:~:text=an%20element%20with%20the%20same%20aria%20role%20as%20these%20elements'%20implicit%20aria%20role%2C%20the%20%3Cheader%3E%20element%20has%20the%20generic%20role%20instead%2C%20and%20is%20no%20longer%20considered%20a%20landmark).

PR #2749, opened 2025-08-29
Refactor convertToDynamoExpressionVars

benmartin-coforma:
### Description
The function `convertDynamoExpressionVars()` was fairly awkward. This PR:
- Removes the 2nd parameter, which turned out to be unused
- Replaces string concatenation with `.join()`
- Replaces `Object.keys(foo) ... foo[k]` with `Object.entries(foo)`
- Replaces repeated object property assignment with `Object.fromEntries()`
- Improves the test for this function
- Improves documentation
- Makes unit tests more specific throughout the rest of the API

Possible future work: I believe it will make sense to remove `convertToDynamoExpressionVars()` (or, as it is now known, `createDynamoUpdateParams()`) entirely. Its job is to convert arbitrary object shapes into Dynamo update commands, but we only actually use it for two (2) object shapes. Both of those would be easier to follow, in my opinion, if we wrote the Dynamo commands directly, hidden away in helper functions called `updateMeasureData()` and `updateCoresetStatus()`. As an example, this is already the pattern we use for `putCombinedRatesToTable()`.

### Related ticket(s)
n/a

---

### How to test

1. Complete a core set
2. Go into one of that core set's measures
3. Make a change to that measure's data
4. Observe that the measure updates, and the core set is kicked back to In Progress.
5. For bonus points, check out the Cloudwatch logs for the `editMeasure` lambda. It will include the actual Dynamo commands for the measure table and coreset table. You may verify that the update expression looks the way we expect it to.

### Notes
n/a

---

### Pre-review checklist

---

#### Review

#### Security


ailZhou (approved):
lgtm, didn't have any issues with saving/loading coresets

PR #2755, opened 2025-09-03
CMDCT-4825 - added html sanitizier to banner

Filtos:
### Description
- Banner descriptions are now being sanitized before being sent to the parser. The sanitization config is setup to match the one in the backend, so now the user will get a preview of exactly what is saved in the DB.
- Changed the `<p>` to a `<span>` so there would be no errors if a user provides a `<ul>` (as a `<ul>` should not be inside a `<p>`)

### Related ticket(s)

CMDCT-4825

---

### How to test
- Login as an admin: https://d1fv14y2db6spl.cloudfront.net/
- Modify the banner to something like:
```
<ul>
<li> list item </li>
</ul>
```
and verify there are no errors in the console

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/Banner/Banner.tsx
+++ b/services/ui-src/src/components/Banner/Banner.tsx
@@ -1,15 +1,27 @@
 import { Alert } from "@cmsgov/design-system";
 import { BannerData } from "types";
 import { parseLabelToHTML } from "utils";
+import DOMPurify from "dompurify";
+
+const domPurifyBannerConfig = {
+  // Only these tags will be allowed through
+  ALLOWED_TAGS: ["ul", "ol", "li", "a", "#text", "strong", "b", "em"],
+  // On those tags, only these attributes are allowed
+  ALLOWED_ATTR: ["href", "alt"],
+  // If a tag is removed, so will all its child elements & text
+  KEEP_CONTENT: false,
+};
 
 export const Banner = ({ bannerData, ...props }: Props) => {
   return (
     <>
       {bannerData && (
         <Alert heading={bannerData.title} {...props}>
-          <p className="ds-c-alert__text">
-            {parseLabelToHTML(bannerData.description)}
-          </p>
+          <span className="ds-c-alert__text">
+            {parseLabelToHTML(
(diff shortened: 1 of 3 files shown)


services/ui-src/yarn.lock
```diff
@@ -4,15 +4,15 @@
 
 "@ampproject/remapping@^2.2.0":
   version "2.3.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.3.0.tgz#ed441b6fa600072520ce18b43d2c8cc8caecc7f4"
+  resolved "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz"
```
cy-by, line 7: Nit: did you install the `dompurify` package via `npm install` instead of `yarn add`?

This isn't a big deal, it just changes a lot of paths that don't seem like they need to change.

Filtos: You're right, thanks for the catch. Fixed.

ailZhou (dismissed):

rocio-desantiago (approved):
Tested and no console errors

services/ui-src/src/components/Banner/Banner.tsx
```diff
@@ -1,15 +1,27 @@
...
 
 export const Banner = ({ bannerData, ...props }: Props) => {
   return (
     <>
       {bannerData && (
         <Alert heading={bannerData.title} {...props}>
-          <p className="ds-c-alert__text">
-            {parseLabelToHTML(bannerData.description)}
-          </p>
+          <span className="ds-c-alert__text">
```
cy-by, line 20: A `<ul>` is not allowed in a `<span>`, I'd suggest using a `<div>`.

From an [HTML validator](https://validator.w3.org/):

QMR UL in Span error

Because a `<span>` expects to have [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) as children.

```suggestion
          <div className="ds-c-alert__text">
```

cy-by (approved):

PR #2763, opened 2025-09-11
[CMDCT-5105] Validation 2021

ailZhou:
### Description

It has been eons since I've started this PR and now it's finally ready for review. Majority of the `validation.ts` has been removed from 2021 measures, only 3 are left as those measures are a bit unique. All the 2021 validations functions are now in `measures/2021/validationTemplate.tsx` [view file here](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/650b92c5994429ccac978c469b298c06b4628a2e/services/ui-src/src/measures/2021/validationTemplate.tsx).

ValidationTemplate.tsx is where every validation function for the measure year is stored to be called. This is the only file that's needed to house all the functions unlike before. It's a bunch of switch/cases that uses the value being passed from data.ts to determine which one to use. The OMS validation functions are in their own switch/case as they run a little bit differently. 

Screenshot 2025-09-19 at 10 55 28 AM

There's also a couple of measures that have unique function calls: AIFHH, IUHH & PCRAD/PCRHH, so they have their own section for OMS.

Screenshot 2025-09-19 at 10 55 14 AM

Inside each `measure/2021/<measure name>/data.ts`, there are two new objects: 

Screenshot 2025-09-19 at 10 40 44 AM

`validations: [ ]` :  This is how we indicate what validation needs to run for the measure when the user clicks validate.
`override: { }` :  I couldn't think of a better way to provide values to override the validation functions parameterers. My longterm goal is to eventually refactor a couple of the validations so that they don't need to call this. I have some suspicious that these functions can be improved.

Other files that were updated:
- **validateSameDenominatorSets/index.ts**: AMM-AD, IET-AD & IET-HH had the `validateSameDenominatorSetsOMS` function stored in their validation.ts file. I migrated it over to the correct validation function file so that the code was in one central location.
- **dataDrivenTools.ts**: turns out, none of the pre-2023 OMS callbacks could be reached (caused by my previous refactoring) so this fixes it. 
- **validationMock.tsx**: needed to be updated for that the mock was able to run the unit tests correctly in `measureTemplate.test.tsx`.

**Measures not switch to using validationTemplate.ts**: These measures are a bit unique and doesn't utilize a data.ts file so I will need to look at refactoring them later or just letting them be.
- CPAAD
- CPCCH
- SSHH

**Auto Complete Measures** - Ignore these when testing
- LBWCH
- LRCDCH
- NCIDDSAD
- PDENTCH

### Related ticket(s)

CMDCT-5105

---

### How to test

Deploy Link: https://d31v90228jijz9.cloudfront.net/

1) Sign into QMR, any state user i.e. [stateuserWA@test.com](mailto:stateuserWA@test.com)
2) Select Reporting Year 2021
3) Randomly choose any measure and start testing validation. You do not need to test every measure, that would be unreasonable. 
4) If no bugs can be see and validation is working, that is good enough.

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/measures/2021/AMRAD/data.ts
+++ b/services/ui-src/src/measures/2021/AMRAD/data.ts
@@ -1,15 +1,19 @@
-import { DataDrivenTypes } from "shared/types";
 import { getCatQualLabels } from "../rateLabelText";
+import { MeasureTemplateData } from "shared/types/MeasureTemplate";
 
 export const { categories, qualifiers } = getCatQualLabels("AMR-AD");
 
-export const data: DataDrivenTypes.PerformanceMeasure = {
-  questionText: [
-    "The percentage of beneficiaries ages 19 to 64 who were identified as having persistent asthma and had a ratio of controller medications to total asthma medications of 0.50 or greater during the measurement year.",
-  ],
-  questionListItems: [],
-  categories,
-  qualifiers,
+export const data: MeasureTemplateData = {
+  type: "HEDIS",
+  coreset: "adult",
+  performanceMeasure: {
+    questionText: [
+      "The percentage of beneficiaries ages 19 to 64 who were identified as having persistent asthma and had a ratio of controller medications to total asthma medications of 0.50 or greater during the measurement year.",
+    ],
+    questionListItems: [],
+    categories,
+    qualifiers,
+  },
   validations: [
     "validateRequiredRadioButtonForCombinedRates",
     "validateAtLeastOneDeviationFieldFilled",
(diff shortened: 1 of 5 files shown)


ailZhou:
> I tried filling out PQI15-AD, and the only three fields I filled out were these:
> 
> Screenshot 2025-09-22 at 11 33 59 AM
> Expected some validation errors to show up but it validated successfully. Tried filling out those same fields for the measure + validating for 2022 and these are the validation errors I'm seeing: Screenshot 2025-09-22 at 11 35 00 AM
> 
> I am assuming these validation errors should be showing up for 2021 as well?

ah shoot, you're right! the functions are getting scrambled when compiled for web, so it's actually running into an error.

Screenshot 2025-09-22 at 12 12 03 PM

PR #2781, opened 2025-09-24
Upgrade vite

gmrabian:
### Description

Saw this deprecation warning for Vite:
`The CJS build of Vite's Node API is deprecated.`

The [troubleshooting guide](https://v6.vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated) indicates we can explicitly declare `type` in our package.json to fix this. We are already using ESM everywhere in the frontend (as dictated by the tsconfig (I think), rather than our package.json). Vite v7 drops support for CJS and thus removes this warning. The [vite changelog](https://github.com/vitejs/vite/blob/main/packages/vite/CHANGELOG.md) indicates no breaking changes we should be worried about. All we need to do is bump our node minor version and we can use vite v7. Worth it to remove one warning? 🤷 

---
### How to test

- All e2e and unit tests pass
- You can run the app locally and deployed build succeeds
- The app behaves as expected

https://d2h41ereaeg3ci.cloudfront.net/

### Notes

`nvm install`
`cd services/ui-src`
`yarn`

---

### Pre-review checklist

---


benmartin-coforma (approved):

rocio-desantiago (approved):
Not seeing anything affected 👍

PR #2985, opened 2026-04-08
[QMR] - Snyk :  Inefficient Algorithmic Complexity

rocio-desantiago:
### Description
This PR focuses on updating minimatch on it's own through resolutions. 
A separate ticket has been created to address other package version upgrades:
[Spike - Update packages to resolve post resolution validation warnings](https://jiraent.cms.gov/browse/CMDCT-5801)

> To be extremely clear:

> [!WARNING]
> If you create a system where you take user input, and use that input as the source of a Regular Expression pattern, in this or any extant glob matcher in JavaScript, you will be pwned.
> 

**Context from Jira ticket**:
For this vulnerability, current version is : 3.1.2 and new version recommendations : 9.0.7,8.0.6,7.4.8,6.2.2,5.1.8,4.2.5,3.1.3,10.2.3

#### Overview
[minimatch](https://www.npmjs.com/package/minimatch) is a minimal matching utility.

Affected versions of this package are vulnerable to Inefficient Algorithmic Complexity via the `matchOne` function. An attacker can cause significant delays in processing and stall the event loop by supplying specially crafted glob patterns containing multiple non-adjacent GLOBSTAR segments.

Remediation
Upgrade `minimatch` to version 3.1.3, 4.2.5, 5.1.8, 6.2.2, 7.4.8, 8.0.6, 9.0.7, 10.2.3 or higher.
References
[GitHub Commit](https://github.com/isaacs/minimatch/commit/0bf499aa45f5059b56809cc3b75ff3eafeb8d748)

### Related ticket(s)
CMDCT-5636

---
### How to test
**Deploy Link:** https://d2seer0curzubk.cloudfront.net/
Nothing should change.
Everything works as-is, as-was.

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,5 +71,10 @@
     "util": "^0.12.5",
     "yargs": "^17.7.2"
   },
+  "resolutions": {
+    "minimatch@^10.1.1": "10.2.5",
+    "minimatch@^3.0.4": "3.1.5",
+    "minimatch@^3.1.1": "3.1.5"
+  },
   "packageManager": "yarn@4.12.0"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5233,22 +5233,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@isaacs/balanced-match@npm:^4.0.1":
-  version: 4.0.1
-  resolution: "@isaacs/balanced-match@npm:4.0.1"
-  checksum: 10c0/7da011805b259ec5c955f01cee903da72ad97c5e6f01ca96197267d3f33103d5b2f8a1af192140f3aa64526c593c8d098ae366c2b11f7f17645d12387c2fd420
-  languageName: node
-  linkType: hard
-
-"@isaacs/brace-expansion@npm:^5.0.0":
-  version: 5.0.0
-  resolution: "@isaacs/brace-expansion@npm:5.0.0"
-  dependencies:
-    "@isaacs/balanced-match": "npm:^4.0.1"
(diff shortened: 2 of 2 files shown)


stacy-coforma (approved):
The only thing I was really wondering is if we could declare both of those 3.x dependencies in the same line, but some Googling has informed me that Yarn doesn't really support that cleanly :| so I think this is fine for now, but we'll need to watch for changes to those versions within the packages using this library (which I'm sure is going to be handled with the bigger rabbit hole ticket(s)) so I am fine with this for the time being

benmartin-coforma:
I approve of these changes.

For the record, I would always rather upgrade our direct dependencies than override version numbers for our transient dependencies, but that was not possible in this case.

I also expect that, if/when our dependencies upgrade to use minimatch versions not specified in our new resolution config, they will upgrade to versions unaffected by this security warning - making our resolution config obsolete, but not re-exposing us to the bug we're trying to avoid.

Even if I am wrong about that,
   1. Snyk will warn us again, and we can update our config again
   2. We won't see any actual impact. Our direct dependencies which ultimately pull in minimatch are `jest`, `aws-cdk-lib`, `tsx`, `babel-jest`, `patch-package`, `sass`, and `vite`. All of these are dev-time tools; none of them are exposed to untrusted input. Therefore our main motivation here not to fix an actual security issue, but just to appease Snyk.

Although we should definitely get rid of some of our extra packages. For example, I know for a fact we do not use `patch-package`, and we could uninstall it without noticing it's gone.

benmartin-coforma (approved):

PR #2986, opened 2026-04-09
CMDCT 5691 - 2026 - PRS-CH/AD - Validation

sopspops:
### Description

* Validation for PRS-CH and PRS-AD
* If the combination rate is greater than the influenza or Tdap rates, than error message should display

### Related ticket(s)

CMDCT-5691

---
### How to test

https://d1ccbk5jn555rm.cloudfront.net/
* Navigate to PRS-CH and PRS-AD
* Fill the form and for this part "Enter a number for the numerator and the denominator. Rate will auto-calculate:" test each cases (for exmaple, Influenza numerator and and denominator - 1, 4, for Tdap - 1, 3, and for combination - 1, 2)
* See Rate error is displayed 

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    const hasLowerOtherRate = otherQualIndices.some((idx) => {
+      const otherRate = parseFloat(ratefields[idx]?.rate ?? "");
+      return !Number.isNaN(otherRate) && qualRate > otherRate;
+    });
+
+    if (hasLowerOtherRate) {
+      errorArray.push({
+        errorLocation: location,
+        errorMessage:
+          "Combination rate cannot be greater than the Influenza or Tdap rates",
```
stacy-coforma: I'm thinking about this error message probably too much so I thought I'll just add my thoughts as a comment here to get them out of my head 😆 

Because the function itself takes in parameters for the indices, I am kind of thinking I'd like this error message to be less hard-coded, because theoretically it could be used for ANY set of indices 🤔 allowing it to accept parameters could lead to the function being reused somewhere else, and then this error message would need to be changed to match whatever the new indices provided are.

Instead I'm wondering if it might make sense to pass the intended error message as a parameter?  So this way if someone wanted to reuse this for another set of indices, they could just provide their own error message and we wouldn't have to worry about a code change to the internals of this function?

Let me know if that makes sense and/or what you think!  👀

sopspops: @stacy-coforma Thank you for reviewing Stacy! and that totally makes sense, do we want to make the entire errorMessage as a parameter like `errorMessage: ${errorMessage}` or something like this?
`return ${highQual} rate cannot be greater than ${lowQual} or ${anotherLowQual} rates`

I've been playing around with the second one, but since the name for each prop is pretty long (for example - Tetanus, Diphtheria Toxoids and Acellular Pertussis (Tdap): Age 21 and Older) and we want to display it as Tdap for this one, so I've been trying to figure it out how to do it!

stacy-coforma: I had like a whole thing typed out and then I lost it in a browser restart 😩 

I think for this specifically it might be most straightforward to pass in the entire `errorMessage`, since the wording does not cleanly map to anything that exists within the `qualifiers` object right now.  If we tried to use the labels/text provided there already, it'd be far wordier than what they were requesting.

I'm happy to defer to anyone who knows the app better and can scheme up a quicker and better mapping, of course.

sopspops: That totally makes sense!! I'll try first but if I feel like it will take me forever, I'd love to defer to anyone who knows the app better hehe. Thank you @stacy-coforma🫶

services/ui-src/src/measures/2026/PRSCH/data.ts
```diff
@@ -71,5 +73,9 @@ export const data: MeasureTemplateData = {
     validateEqualCategoryDenominatorsPM: {
       qualifiers,
     },
+    validateOneQualRateLessThanOrEqualToOtherQualRates: {
+      qualIndex: 2,
+      otherQualIndices: [0, 1],
+    },
```
stacy-coforma, line 81: credit goes to Ben for tossing around ideas but what do you think about this:

```suggestion
    validateOneQualRateLessThanOrEqualToOtherQualRates: {
      qualIndex: 2,
      otherQualIndices: [0, 1],
      errorMessage: "Combination rate cannot be greater than the Influenza or Tdap rates"
    },
```

Another thing Ben suggested which I'm on board with is using the `id` field on the qualifier so that the lookups are tied to the objects themselves rather than the positions in the qualifiers array; I think this makes a lot of sense and takes away some of the risk of any ordering changes in the future affecting the behavior.  Theoretically, if we need to make a change there we could also make it here so the indices match up correctly, but I could see a future where we forget that this validator is here 😅 that particular hill is not one I'm willing to die on though.  Functionally the code as written is doing what the ticket asks it to do!

sopspops: Thank you! It totally makes sense! I made errorMessage as a prop and added Ids fields. So... ready for re-review🫶 @stacy-coforma

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.ts
```diff
@@ -8,18 +8,56 @@ import {
   convertOmsDataToRateArray,
 } from "../dataDrivenTools";
 
+const DEFAULT_ERROR_MESSAGE =
```
stacy-coforma, line 11: this was a good move 👍

services/ui-src/src/measures/2026/validationTemplate.tsx
```diff
@@ -153,6 +153,19 @@ const omsValidations = (func: ValidationFunction, PMD: MeasureTemplateData) => {
         PMD.override?.validateOneQualRateHigherThanOtherQual?.higherIndex,
         PMD.override?.validateOneQualRateHigherThanOtherQual?.lowerIndex
       );
+    case "validateOneQualRateLessThanOrEqualToOtherQualRatesOMS":
+      return GV.validateOneQualRateLessThanOrEqualToOtherQualRatesOMS(
+        PMD.override?.validateOneQualRateLessThanOrEqualToOtherQualRates
+          ?.qualIndex,
+        PMD.override?.validateOneQualRateLessThanOrEqualToOtherQualRates
+          ?.otherQualIndices,
```
stacy-coforma: Can these now be removed?

services/ui-src/src/measures/2026/validationTemplate.tsx
```diff
@@ -377,6 +390,21 @@ export const validationTemplate = (
...
           PMD.override?.validateOneQualRateHigherThanOtherQual?.lowerIndex
         );
+      case "validateOneQualRateLessThanOrEqualToOtherQualRatesPM":
+        return GV.validateOneQualRateLessThanOrEqualToOtherQualRatesPM(
+          data,
+          PMD.performanceMeasure,
+          PMD.override?.validateOneQualRateLessThanOrEqualToOtherQualRates
+            ?.qualIndex,
+          PMD.override?.validateOneQualRateLessThanOrEqualToOtherQualRates
+            ?.otherQualIndices,
```
stacy-coforma: Can these now be removed?

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.test.ts
```diff
@@ -0,0 +1,228 @@
...
+      const errors = validateOneQualRateLessThanOrEqualToOtherQualRatesPM(
+        data,
+        {
+          categories,
+          qualifiers,
+        }
+      );
+      expect(errors).toHaveLength(1);
+      expect(errors[0].errorMessage).toBe(
+        "Combination rate cannot be greater than the Influenza or Tdap rates"
```
stacy-coforma, line 124: May be worth throwing this as a `const` at the top of the file since it'll be validated against so much in these tests (absolute nit, non blocking)

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return _validation({
+      qualifiers,
+      location: `Optional Measure Stratification: ${locationDictionary(label)}`,
+      qualIndex: resolveQualifierIndex(qualifiers, qualId, qualIndex),
+      otherQualIndices: resolveQualifierIndices(
+        qualifiers,
+        otherQualIds,
+        otherQualIndices
+      ),
```
stacy-coforma: Keeping the indices in the `validationTemplate` makes sense now that I see this portion 😅

Ideally I think it would be best, if we are going to use the `id` instead of the index of the element within the array, to remove the need to pass the index as well.

I do also see now that I am looking further into it that the `FormRateField` does not really include a good way to get to the `id`.  However running locally I can see that we receive a `uid` field on those objects, so I'm thinking we could add `uid` as an optional field on that interface so that we can grab it from within the validation, and then we wouldn't necessarily need to pass in the index.

Or, if that is not possible, I think it's fine to just go back to relying on the index exclusively.  I would prefer the `id` field but not if it's going to create two things to maintain instead of just one 😅 sorry about the kind of runaround on this!

sopspops: Thank you so much for going back and forth on this PR with me, and thorough review🥹 I made changes, how does it look now?

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.ts
```diff
@@ -0,0 +1,155 @@
...
+) => {
+  if (qualIds.length > 0) {
+    const matchedFields = qualIds
+      .map((qualId) =>
+        ratefields.find(
+          (ratefield) =>
+            ratefield?.uid === qualId ||
+            getQualifierIdFromUid(ratefield?.uid) === qualId
+        )
+      )
```
stacy-coforma: just a small suggestion for reusing the function above

```suggestion
      .map((qualId) => findRateFieldByQualifier(ratefields, qualId))
```

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        },
+        validationOptions
+      );
+      expect(errors).toHaveLength(1);
+      expect(errors[0].errorMessage).toBe(
+        "Combination rate cannot be greater than the Influenza or Tdap rates"
+      );
+    });
+
+    it("returns no error for single category label", () => {
```
stacy-coforma: naming mismatch - from looking at this it appears to be doing what we expect but the name suggests it shouldn't error.  (so I think renaming is the move unless I've misread something in here where it shouldn't actually error?)

```suggestion
    it("returns error for single category label when combination rate is greater than another rate", () => {
```

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        validationOptions
+      )({
+        ...baseOMSInfo,
+        rateData,
+      });
+      expect(errors).toHaveLength(1);
+      expect(errors[0].errorMessage).toBe(
+        "Combination rate cannot be greater than the Influenza or Tdap rates"
+      );
+    });
```
stacy-coforma: Don't think we need this one anymore; it's identical to a test a couple tests above it

```suggestion
    });
```

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        {
+          qualId: "Combination",
+          otherQualIds: ["Influenza", "Tdap"],
+        }
+      );
+      expect(errors).toHaveLength(1);
+      expect(errors[0].errorMessage).toBe(
+        "Combination rate cannot be greater than the Influenza or Tdap rates"
+      );
+    });
```
stacy-coforma: this can get moved up to the other block with the other PM tests

stacy-coforma (commented):
looks good and functionally still works great :) I added some pretty minor comments mostly in the test file, and then I think we will be good to go!

services/ui-src/src/shared/globalValidations/validateOneQualRateLessThanOrEqualToOtherQualRates/index.test.ts
sopspops: weee thank you Stacy! Test fixed and function cleaned up- ready for re-review! @stacy-coforma

stacy-coforma (approved):
thank you for going along that id vs index journey with me 😅 and thank you for getting the changes in so fast!

PR #2997, opened 2026-04-27
Fix branch name in README

benmartin-coforma:
### Description
Just a small correction here.


stacy-coforma (approved):
good catch! in my brain I had corrected that 😅

PR #3003, opened 2026-05-04
2026 Child Core Set Updates

sopspops:
### Description
* Child Core Set Measure Updates
* CHL-CH, CIS-CH, FUH-CH, FUM-CH, IMA-CH, OEVP-CH, PPC2-CH, PDS-CH,PRS-CH, WCC-CH

### Related ticket(s)
CMDCT-5863

---
### How to test
https://d1bub1gpkw7as4.cloudfront.net/
* Navigate to each measure and see the updates

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/measures/measureDescriptions.ts
```diff
@@ -482,7 +482,7 @@ export const measureDescriptions: MeasureList = {
     "LRCD-CH": "Low-Risk Cesarean Delivery: Under Age 20",
     "LSC-CH": "Lead Screening in Children",
     "OEV-CH": "Oral Evaluation, Dental Services",
-    "OEVP-CH": "Oral Evaluation During Pregnancy: Ages 15 to 20",
+    "OEVP-CH": "Oral Evaluation During Pregnancy: Ages 15 through 20",
```
stacy-coforma: Was this one unintentional? Looks like it is applied to the 2025 list?

services/ui-src/src/measures/2026/FUHCH/data.ts
```diff
@@ -8,7 +8,7 @@ export const data: MeasureTemplateData = {
   coreset: "child",
   performanceMeasure: {
     questionText: [
-      "Percentage of discharges for beneficiaries ages 6 to 17 who were hospitalized for treatment of selected mental illness or intentional self-harm diagnoses and who had a follow-up visit with a mental health provider. Two rates are reported:",
+      "Percentage of discharges for beneficiaries ages 6 to 17 who were hospitalized for principal diagnosis of mental illness, or any diagnosis of intentional self-harm, and had a follow-up service. Two rates are reported:",
```
benmartin-coforma:
```suggestion
      "Percentage of discharges for beneficiaries ages 6 to 17 who were hospitalized for a principal diagnosis of mental illness, or any diagnosis of intentional self-harm, and had a follow-up service. Two rates are reported:",
```

benmartin-coforma (commented):
One typo here, plus Stacy's question. It does seem like maybe we should change the OEVP-CH measure description? But probably only for 2026.

stacy-coforma:
> One typo here, plus Stacy's question. It does seem like maybe we should change the OEVP-CH measure description? But probably only for 2026.

I am curious on this because the title is not listed as a change in the spec doc but would make sense to do given the other changes.  Could be a question to save for Cam when he returns 🤔 

EDIT: jkjk, the description IS changed, the TITLE is what is not called out in the spec doc as a change 😵‍💫 words are hard for me sometimes.

services/ui-src/src/measures/measureDescriptions.ts
sopspops: ooope yup my mistake! Thank you!!

services/ui-src/src/measures/2026/FUHCH/data.ts
sopspops: Thank you!!

sopspops:
> > One typo here, plus Stacy's question. It does seem like maybe we should change the OEVP-CH measure description? But probably only for 2026.
> 
> I am curious on this because the title is not listed as a change in the spec doc but would make sense to do given the other changes. Could be a question to save for Cam when he returns 🤔
> 
> EDIT: jkjk, the description IS changed, the TITLE is what is not called out in the spec doc as a change 😵‍💫 words are hard for me sometimes.

@stacy-coforma @benmartin-coforma Thank you for catching typos and errors! All fixed now!

stacy-coforma (approved):
I am good with this, but I think we ought to flag to Cam the title change for OEVP-CH to make sure that's going to be fine with the BOs 👀 based on the other updates it does seem to make sense, but just to get confirmation.

PR #3013, opened 2026-05-27
2026 - Other Data Source Updates

sopspops:
### Description

* Several contents updated related to Other Data Source, 2026 only

### Related ticket(s)

CMDCT-6061

---
### How to test

https://d2fob2ca9olnrc.cloudfront.net/
* For 2026- see some contents updated related to Other Data Source

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/measures/2026/COLHH/data.ts
+++ b/services/ui-src/src/measures/2026/COLHH/data.ts
@@ -16,7 +16,7 @@ export const data: MeasureTemplateData = {
   },
   dataSource: {
     optionsLabel:
-      "If reporting entities (e.g., health plans) used different data collection methods, please select all that are applicable below. If your state followed the 2026 Core Set technical specifications and used administrative data only, select ECDS > Administrative.",
+      "If reporting entities (e.g., health plans) used different data collection methods or sources, please select all that are applicable below. If your state followed the 2026 Core Set technical specifications and used administrative data only, select ECDS > Administrative.",
     options: [
       {
         value: DC.ELECTRONIC_CLINIC_DATA_SYSTEMS,
--- a/services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
+++ b/services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
@@ -34,7 +34,7 @@ export const DataSource = ({ type }: DataSourceProps) => {
                 name={DC.DATA_SOURCE_CAHPS_VERSION_OTHER}
                 label={
                   <>
-                    Describe the data source (
+                    Describe the data collection method or data source (
                     <em>
                       text in this field is included in publicly-reported
                       state-specific comments
--- a/services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
+++ b/services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
@@ -34,7 +34,7 @@ export const DataSource = ({ type }: DataSourceProps) => {
                 name={DC.DATA_SOURCE_CAHPS_VERSION_OTHER}
                 label={
                   <>
-                    Describe the data collection method or data source (
+                    Describe the data source (
                     <em>
                       text in this field is included in publicly-reported
                       state-specific comments
--- a/services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
+++ b/services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
@@ -31,11 +31,13 @@ export const validateAtLeastOneDataSourceType = (
           errorLocation: dataSourceLabel,
           errorMessage:
             errorMessage ??
-            `Please describe the ${label}${
-              !label.includes("Source") && !label.includes("Method")
-                ? " Source"
-                : ""
-            }`,
+            (featuresByYear.useDataCollectionMethod
+              ? "Please describe the Other Data Collection Method or Data Source"
+              : `Please describe the ${label}${
+                  !label.includes("Source") && !label.includes("Method")
+                    ? " Source"
+                    : ""
+                }`),
         };
       })
     );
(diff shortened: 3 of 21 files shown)


services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
```diff
@@ -31,11 +31,13 @@ export const validateAtLeastOneDataSourceType = (
           errorLocation: dataSourceLabel,
           errorMessage:
             errorMessage ??
-            `Please describe the ${label}${
-              !label.includes("Source") && !label.includes("Method")
-                ? " Source"
-                : ""
-            }`,
+            (featuresByYear.useDataCollectionMethod
+              ? "Please describe the Other Data Collection Method or Data Source"
```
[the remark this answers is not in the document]
stacy-coforma: It took me a long time to parse this and I think it's not a good suggestion 😅 but I think it is rooted in an actual thing to address - I think we want to check both `featuresByYear.useDataCollectionMethod` and `lookupKey === DataSource.OtherDataSource` here, because while I was testing I randomly chose measure CCW-AD which has an "Administrative Other" choice that also requires a comment.  I'm not sure if that should fall into the category of needing the same message or not; might be worth checking with Cam on that particular question.

I'm not sure if there are any other selections that could require a comment, but even having one that differs probably makes it worth specifying to exactly the options we know this should be applicable to.

services/ui-src/src/measures/2026/COLHH/data.ts
stacy-coforma, line 1: This looks good but when matching against the ticket, I think COL-AD was missed!  I can't comment on that file since it's not part of the set of changes 🙈

services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
```diff
@@ -34,7 +34,7 @@ export const DataSource = ({ type }: DataSourceProps) => {
                 name={DC.DATA_SOURCE_CAHPS_VERSION_OTHER}
                 label={
                   <>
-                    Describe the data collection method or data source (
+                    Describe the data source (
```
stacy-coforma: I'm not sure if this one needs to change?  In the ticket I see it asking for this for Administrative Other text, but this one doesn't appear to be related to Administrative Other?  Could be another one for Cam to verify maybe 🤔

services/ui-src/src/measures/2026/COLHH/data.ts
sopspops: oooh thank you for catching that!!

services/ui-src/src/measures/2026/MSCAD/questions/DataSource.tsx
sopspops: Now i'm checking it again, I don't think we need this as this falls under "other"- my mistake!!

services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
```diff
@@ -31,11 +31,13 @@ export const validateAtLeastOneDataSourceType = (
           errorLocation: dataSourceLabel,
           errorMessage:
             errorMessage ??
-            `Please describe the ${label}${
-              !label.includes("Source") && !label.includes("Method")
-                ? " Source"
-                : ""
-            }`,
+            (featuresByYear.useDataCollectionMethod
```
stacy-coforma:
```suggestion
            (featuresByYear.useDataCollectionMethod && lookupKey === DataSource.Other
```

I think we'll want this to also check that our lookup key is the `OtherDataSource` key so that we are selectively showing this message only to the ones marked "Other"

(Edited for wrongness of `Other` key)

services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
```diff
@@ -31,11 +31,13 @@ export const validateAtLeastOneDataSourceType = (
...
+              ? "Please describe the Other Data Collection Method or Data Source"
+              : `Please describe the ${label}${
+                  !label.includes("Source") && !label.includes("Method")
+                    ? " Source"
+                    : ""
+                }`),
         };
```
benmartin-coforma, line 46: In addition to the LLM's correct note (that we cannot ignore the value of `label` in 2026), the ternary operators are getting kind of deep here. It might be easier to read if you move some of the logic up into `if` statements to set some new variables.

benmartin-coforma:
Here are some example validation messages. When multiple data sources needing explanation are selected:
image

In 2025 we would call them out separately:
image

Whereas with the current logic in this branch, in 2026, we do not:
image

I captured those screenshots from this branch's deployment environment, for the WCC-CH measure, in the state of Massachusets (that is, as `stateuser1@test.com`). https://d2fob2ca9olnrc.cloudfront.net/MA/2026/CCSM/WCC-CH

sopspops:
@benmartin-coforma 
Screenshot 2026-06-01 at 1 49 07 PM

Oooh I didn't even notice this! Thank you Ben!! It is fixed now!

services/ui-src/src/shared/globalValidations/validateAtLeastOneDataSourceType/index.ts
```diff
@@ -27,17 +27,24 @@ export const validateAtLeastOneDataSourceType = (
...
+          lookupKey === DataSource.Other
+        ) {
+          generatedMessage =
+            "Please describe the Other Data Collection Method or Data Source";
+        } else {
+          const shouldAppendSource =
+            !label.includes("Source") && !label.includes("Method");
+          const sourceSuffix = shouldAppendSource ? " Source" : "";
+          generatedMessage = `Please describe the ${label}${sourceSuffix}`;
+        }
```
stacy-coforma: I might consider modifying this to be a function to create the message instead, because I like being able to `return "xyz"` instead of if/else-ing, but this is merely a suggestion 😅

sopspops: That makes sense!! Thank you for the suggestoin :) Just updated it and It looks much simpler as well!

stacy-coforma (approved):
Thank you for making so many adjustments 🙏 I feel like we all learned a lot about copilot review with this one 🤣 thank you for being so patient!

PR #3020, opened 2026-06-03
[CMDCT-5998] Security Fix Part 2: More quick fix updates

ntsummers1:
### Description

This is part 2 with some more easy fixes for a few more security vulnerabilities flagged by dependabot. Most of the work is just bumping outdated packages to their latest safe versions across the codebase. I then went through and removed dependencies that are no longer in use or were actually conflicting with other dependencies. The main call out here though is JSX. JSX types got added to React in React 19, so I needed to let all of our files know about that change. So 59 files changed, but 1 line in each of them. Not too bad!

### Related ticket(s)

CMDCT-5998

---
### How to test

This is a tried and true load up the application and play around. Everything should still work as expected. Nothing should seem off. Please report anything you find weird or seems incorrect!

#### App-api
"aws4": Removed
"uuid": "^7.0.3" -> "11.1.1"
"@types/aws4": Removed

#### Ui-Src
"aws-amplify": "~6.16.2" -> "~6.17.0"
"base-64": Removed
"launchdarkly-react-client-sdk": "^3.9.2",
"node-fetch": Removed
"node-forge": Removed
"react-helmet-async": "^2.0.5" (Moved from root package to ui-src package)
"react-router-dom": "6.27.0" -> "6.30.0"
"uuid": "^8.3.2" -> "^11.1.1"
"@types/base-64": Removed
"@types/react-router-dom": Removed
"@types/uuid": Removed,
"patch-package": "^6.4.7 -> "^8.0.1"
"postinstall-postinstall": Removed

#### Root
"jsdom": (Removed, already set in app-api where its used)
"@types/dotenv": Removed
"@types/react": Removed
"util": Removed

---
### Pre-review checklist


stacy-coforma (approved):
I've randomly clicked around a bunch!  Everything looks like it's working fine.  I am curious on why `JSX` had to be imported from `react` into so many places 💀 I don't know that I've ever seen that before, but I also haven't worked in React 19 before so possibly something there?  Willing to learn if you know the answer :)

ntsummers1:
> I've randomly clicked around a bunch! Everything looks like it's working fine. I am curious on why `JSX` had to be imported from `react` into so many places 💀 I don't know that I've ever seen that before, but I also haven't worked in React 19 before so possibly something there? Willing to learn if you know the answer :)

I have answers 😄! The reason JSX was working before is because it was using the `@types/react` package. The project then knew about the type JSX globally. I removed that package because React v19 added the JSX type. That does come with the downside of having to either import JSX into every component, or adding it to a global type definitions and importing it into the project that way. We're already importing react into many of our components so I figured that was a cleaner way to clearly define where that type is coming from, but if you'd like I can put it in a global type def and have it read from there.

I should also mention, that `@types/react` package deprecated JSX in version 18, so we needed to remove the package and import it through react to move forward.

PR #3024, opened 2026-06-05
2026 - Foster Care/Expanded Medicaid

sopspops:
### Description

* Add Foster Care and Expanded Medicaid section for Adult and Child

### Related ticket(s)

CMDCT-6058

---
### How to test

https://d3o7jll49b21f4.cloudfront.net/
* Navigate to Child coreset (Medicaid) and see Foster Care is added
* Navigate to Adult coreset (Medicaid) and see Medicaid Expansion is added 

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
+          label: "Adult group – Full expansion",
+          helperText:
+            "Adults enrolled in Medicaid under the adult group at section 1902(a)(10)(A)(i )(VIII) of the Act, and in regulation at 42 C.F.R. § 435.119[1] or in a section 1115 demonstration that provides eligibility for the entire adult group population.",
+        },
+        {
+          id: "medicaid-expansion-non-expansion-beneficiaries",
+          label: "Medicaid-Non-expansion beneficiaries",
+          helperText:
+            "Includes Adults enrolled in Medicaid that are not in the Adult group – Full expansion, as defined above.",
+        },
```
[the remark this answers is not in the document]
stacy-coforma: I am gathering from all of this that copilot thinks hyphens are a problem lol.

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
+      label: "Medicaid Expansion",
+      aggregateTitle: "expansion group",
+      helperText:
+        "Skip the Medicaid expansion stratification if your state has not fully expanded Medicaid for the adult population.",
+      options: [
+        {
+          id: "adult-group-full-expansion",
+          label: "Adult group – Full expansion",
+          helperText:
+            "Adults enrolled in Medicaid under the adult group at section 1902(a)(10)(A)(i )(VIII) of the Act, and in regulation at 42 C.F.R. § 435.119[1] or in a section 1115 demonstration that provides eligibility for the entire adult group population.",
```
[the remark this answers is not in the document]
stacy-coforma: this may be legit

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
 
-const omb1997 = () => {
+const strat2026 = (coreSetId?: string): OmsNode[] => {
+  const data: OmsNode[] = [...omb2024()];
+
+  // Foster Care: Child Medicaid + Health Home only
+  if (coreSetId?.startsWith("HHCS") || coreSetId === "CCSM") {
+    data.push({
+      id: "foster-care",
+      label: "Foster Care",
```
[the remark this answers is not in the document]
stacy-coforma: I do think the repeated criticisms of the hyphenations in here is pretty silly, but I am also noticing that the other data objects in this file are using possibly randomized (?) 6-letter IDs for their ID fields.  We should probably align with that, though I don't know if those IDs are meant to map to something specific or if they were just randomly selected, so we'll need to do a bit of discovery there.

services/ui-src/src/shared/commonQuestions/MeasureStratification/index.tsx
```diff
@@ -103,6 +104,7 @@ export const MeasureStrat = (props: Types.OMSProps) => {
   const labels: any = useContext(SharedContext);
   const year = labels.year;
   const { coreset } = props;
+  const { coreSetId } = useParams();
```
benmartin-coforma, line 107: Do we need the new `useParams()` call here, or does `coreSetId` match the `coreset` prop we already have?

services/ui-src/src/shared/commonQuestions/MeasureStratification/omsNodeBuilder.tsx
```diff
@@ -170,12 +171,22 @@ export const TopLevelOmsChildren = (props: CheckboxChildrenProps) => {
 
   return (
     <CUI.Box key={`${props.name}.topLevelCheckbox`}>
+      {props.helperText && (
+        <CUI.Text fontSize="sm" color="#5A5A5A" mb={4}>
```
benmartin-coforma: Most of the other places in the app that set a color use a variable like `gray.100` or `gray.500` instead of a hex code. Is it possible that one of those variables could give us the correct shade here?

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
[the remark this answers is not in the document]
benmartin-coforma: QMR does weird stuff with its IDs. Someday maybe we can pry off the rest of the `cleanString()` barnacles, but for now the best course of action is to choose IDs that would pass through `cleanString()` unchanged.

That's why the rest of the (non-legacy) OMS node IDs were generated by https://shortunique.id/ - that's probably what we should do for these new IDs too.

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/index.tsx
```diff
@@ -84,7 +90,9 @@ export const OptionalMeasureStrat = ({
   if (IUHHPerformanceMeasureArray || AIFHHPerformanceMeasureArray)
     performanceMeasureArray = undefined;
 
-  const omsData = data ?? OMSData(year, coreset === "adult");
+  const { coreSetId } = useParams();
+  const omsData =
+    data ?? OMSData(year, coreset === "adult", undefined, coreSetId);
```
benmartin-coforma, line 95: It feels redundant to pass both `coreset === "adult"` and `coreSetId` here. Can the `OMSData()` function be changed so that it only needs one of these arguments?

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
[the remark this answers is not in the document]
stacy-coforma: @benmartin-coforma added some context to this in [another comment](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/pull/3024#discussion_r3382201071), which reinforces for me that yes we should use the same strategy to create the IDs for these fields as well. 👍

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
+      options: [
+        {
+          id: "in-foster-care",
+          label: "In foster care during the measurement period",
+        },
```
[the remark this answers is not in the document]
stacy-coforma: resolving this, same ID issue as described in other comments

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
+      addMore: false,
+    });
+  }
+
+  // Medicaid Expansion: Adult Medicaid + Health Home only
+  if (coreSetId?.startsWith("HHCS") || coreSetId === "ACSM") {
+    data.push({
+      id: "medicaid-expansion",
+      label: "Medicaid Expansion",
+      aggregateTitle: "expansion group",
```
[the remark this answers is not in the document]
stacy-coforma: resolving this, same ID issue as described in other comments

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -189,7 +198,58 @@ const omb2024 = () => {
...
+      label: "Foster Care",
+      options: [
+        {
+          id: "in-foster-care",
+          label: "In foster care during the measurement period",
+        },
+        {
+          id: "not-in-foster-care",
+          label: "Not in foster care during the measurement period",
+        },
```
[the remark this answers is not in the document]
stacy-coforma: resolving this, same ID issue as described in other comments

services/ui-src/src/shared/commonQuestions/MeasureStratification/omsNodeBuilder.tsx
sopspops: So I tried them and they are much lighter or doesn't match the shade😭

services/ui-src/src/shared/commonQuestions/MeasureStratification/index.tsx
sopspops: I think we still need coreSetId as these new rules depend on exact id for foster care and expanded medicaid!

sopspops:
@stacy-coforma @benmartin-coforma Thank you for reviewing and the feedback! All the feedback is addressed/applied and ready for re-review!!

services/ui-src/src/shared/commonQuestions/MeasureStratification/omsNodeBuilder.tsx
stacy-coforma: Possibly `gray.600`? I'm looking at a couple things:

- [ChakraUI's Colors definitions for v2](https://v2.chakra-ui.com/docs/styled-system/theme#colors) which to be fair might not match ours 100% due to being on a slightly older version than the current 2.10.10
- Our [colors.ts](https://github.com/Enterprise-CMCS/macpro-mdct-qmr/blob/main/services/ui-src/src/styles/foundations/colors.ts) file which appears to have a couple of overrides

Full disclosure, I have not worked with ChakraUI before and am making assumptions here, but would it make sense to either just use the built-in value for `600` if it is close enough, or to add `#5A5A5A` as an override for `gray`'s `600`?  I don't know how much reuse we might want to get out of this value so I'm fine either way, just offering some thoughts 🙂

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
```diff
@@ -2,9 +2,18 @@ import { OmsNode } from "shared/types";
 
 export const OMSData = (
```
stacy-coforma, line 16: I may suggest leaving a JSDOC or comment here explaining this change, because to me, modifying the second parameter to be `coreSetContext` (string or boolean) instead of `adultMeasure` (boolean only) was quite confusing.  I'm not sure if it makes sense to do it this way versus use a separate property altogether to avoid the confusion for the (many) existing functions that call this already.

services/ui-src/src/shared/commonQuestions/MeasureStratification/omsNodeBuilder.tsx
sopspops: Ooooh that sounds good!!! I'll over write gray.600!! thank you Stacy🥹

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
benmartin-coforma: Oof, yeah. I didn't realize that this function was called so many places; I thought it would be possible to make a clean swap to always pass the full `coreSetId` in place of the boolean.

But a polymorphic `string | boolean` parameter isn't ideal. In theory we could go around and adjust all of the places this function is called (in order to make them pass `"ACS"` instead of `true`), but that would be out of scope for this PR. There may be potential for future refactoring here; it is very strange that we have one function called in so many different ways... Perhaps we could break it somehow?

But again, I don't think that level of effort is warranted for this PR, and I don't think my suggestion made things any better. Four parameters with consistent usage is better than three parameters that may sometimes change their type.

services/ui-src/src/shared/commonQuestions/MeasureStratification/index.tsx
benmartin-coforma: Doesn't the `coreset` prop have the exact ID?

stacy-coforma: coreset appears to just be `adult` or `child` which is surprising to me 🤔

benmartin-coforma: Wow, that's very strange. Maybe we should change it someday; there's no point in having the parent component pass down a prop that contains strictly less than what this component (1) can get from the URL params and (2) needs, in order to render itself.

But, given how the prop works today, this change is fine & good.

services/ui-src/src/shared/commonQuestions/OptionalMeasureStrat/data.ts
benmartin-coforma: I went ahead an reverted this parameter change. My mistake, and sorry for the runaround!

stacy-coforma (approved):
Thank you @benmartin-coforma for making this adjustment!  And great work on everything @sopspops, we just talked about it and determined we should simplify the logic by separating the variables! 🙏

benmartin-coforma (approved):

PR #3038, opened 2026-06-15
[CMDCT-5998] More quick dependency updates 

ntsummers1:
### Description

This PR upgrades Cypress, ESBuild, TSX, and Vite. It also cleans up the lock file for things like pico match and tar. I believe it should get most, if not all, of the high/critical severity security updates. I would love to get this in and see what's left!

### Related ticket(s)

CMDCT-5998

---
### How to test

This is a tried and true load up the application and play around. Everything should still work as expected. Nothing should seem off. Please report anything you find weird or seems incorrect!

### Notes

"aws-cdk": "~2.1112.0" >> "~2.1127.0",
"aws-cdk-lib": "~2.256.0" >> "~2.259.0",
cypress": ^13" >> "^15",
"@types/node": "^24.0.12" >> "^22.19.0" (Should be aligned with the current Node version),
"tsx": 4.20.3" >> "^4.22.4",
"esbuild": >> "^0.28.1",
"@vitejs/plugin-react": "^4.3.1" >> "^6.0.2",
"vite": ^7.3.2" >> "^8.0.16",

---
### Pre-review checklist


benmartin-coforma (dismissed):
Should be fine. ✅

services/app-api/package.json
```diff
@@ -69,8 +69,27 @@
...
+              }
+            ]
+          ]
+        }
+      ],
+      "^.+\\.tsx?$": "ts-jest"
     },
+    "transformIgnorePatterns": [
+      "/node_modules/(?!(@asamuzakjp/|@bramus/|@csstools/|@exodus/bytes/|css-tree/|entities/|lru-cache/|parse5/))"
+    ],
```
ntsummers1, line 92: This was added because jsdom has deps that call esm packages. We actually could get rid of this addition and jsdom by switching to the sanitize-html package we all ready have installed. Not a big lift but does touch prod facing code so i optd to do this for now.

stacy-coforma (dismissed):
poked around in the deployed env and all seems to be working fine, though I found what i think might be a preexisting bug but is unrelated to your changes so approving this :)

.github/scanResources/package.json
```diff
@@ -21,10 +21,10 @@
     "@aws-sdk/client-wafv2": "~3.1058.0",
     "@octokit/rest": "^22.0.1",
     "@types/adm-zip": "^0.5.7",
-    "@types/node": "^24.0.12",
+    "@types/node": "^25.9.3",
```
benmartin-coforma: This doesn't seem right. We're still running Node 22, according to the `.nvmrc`.

services/app-api/package.json
benmartin-coforma: This should also be obviated when we switch from `jest` to `vitest` 👀

services/ui-src/package.json
```diff
@@ -54,11 +54,12 @@
     "@types/file-saver": "^2.0.5",
     "@types/jest": "^30.0.0",
     "@types/jest-axe": "^3.5.9",
+    "@types/node": "^25.9.3",
```
benmartin-coforma: Same comment as above - it feels like this should be `22.something`

tests/cypress/package.json
```diff
@@ -12,8 +12,8 @@
     "test": "concurrently --kill-others \"yarn start\" \"yarn cypress open\""
   },
   "devDependencies": {
-    "concurrently": "^8.2.0",
-    "cypress": "^13",
+    "concurrently": "^10.0.3",
```
benmartin-coforma, line 15: I am very much looking forward to getting rid of `concurrently`. I don't believe we have a strict need for it now, so we _could_ scrap it today. But I think it will also disappear as a matter of course when we switch from Cypress to Playwright; our Playwright implementations in other apps have not been pulling it in.

package.json
```diff
@@ -53,24 +53,27 @@
     "@octokit/auth-action": "^6.0.1",
     "@octokit/rest": "^22.0.0",
     "@types/flat": "^5.0.5",
-    "@types/node": "^22.18.13",
+    "@types/node": "^25.9.3",
```
benmartin-coforma: Same comment here for Node 22.

.github/scanResources/package.json
ntsummers1: I didn't know this! I'll see what I can do to clean this up. Lemme play around a little

tests/cypress/package.json
ntsummers1: I think that might be the right play (Do it when we swap from cypress to playwright)! I would actually love to go ahead and delete as much as I can. But i fear this PR has already bloated above what I wanted it to originally be.

.github/scanResources/package.json
ntsummers1: Looks like it slotted in nicely. No errors from what I can tell!

stacy-coforma (approved):
good catch on the node version @benmartin-coforma 🌟

benmartin-coforma (approved):

PR #3043, opened 2026-06-17
[CMDCT-6158] Remove StatusOfData questions for 2026

ntsummers1:
### Description

CMS would like to remove the Status of Data reported question across the Child (both Medicaid and Separate CHIP), Adult (both Medicaid and Separate CHIP), and Health Home measure templates. This question is found in all measures except for CPC-CH, CPA-AD and the non-reportable measures (LBW-CH, LRCD-CH, LRCD-AD).

### Related ticket(s)

CMDCT-6158

---
### How to test

Important Step: Sign in as stateuser3 (Want something with Health Home core set)

#### Confirm the question is gone from a standard measure (2026)
  1. Log in as a state user and navigate to a 2026 Child or Adult core set
  2. Open any standard measure (e.g., AMR-CH or AMR-AD)
  3. Under the Reporting question, select "Yes, I am reporting"
  4. Scroll through the form — confirm "Status of Data Reported" header and its radio question ("What is the status of the data being reported?") do not appear anywhere on the page

#### Confirm the question is gone from custom-template measures (2026)

Repeat steps 1–4 from above for each of these measures, which have their own templates:
- PCR-AD (Adult core set)
- PCR-HH (Health Home core set)
- MSC-AD (Adult core set)
- SS-1-HH (Health Home core set)

#### Confirm the question still appears in prior years

1. Switch to a 2025 core set
2. Open any standard measure and select "Yes, I am reporting"
3. Confirm "Status of Data Reported" and its radio question do appear

---
### Pre-review checklist

---
#### Review


services/ui-src/src/measures/2026/MSCAD/index.tsx
```diff
@@ -36,7 +36,6 @@ export const MSCAD = ({
           <>
             {data[DC.DID_REPORT] !== DC.NO && (
               <>
-                <CMQ.StatusOfData />
```
ntsummers1, line 39: Newbie question: The notes in this ticket mention that "measures/2026/measuretemplate.tsx is where the change should be made." While I've made the change there, shouldn't it also be removed from these 4 measures as well?

benmartin-coforma: Yes, excellent catch.

stacy-coforma (dismissed):
I did not expect Copilot Review to have so many thoughts on this 😅 

I can see where it is coming from for some of the negative assertion requests, and leave it to you if you'd like to make some/all of the changes it recommends.

Functionally I tested it in the deployed environment and it looks to be working as intended!

ntsummers1:
Updated some of the tests with an expectation of status not being found. I think I actually disagree with all of the copilot suggestions but I can see some possible value from a couple of them specifically so I've added those in.

stacy-coforma (approved):
Thanks for taking a look at the comments, I agree with your assessment 🙂

PR #3047, opened 2026-06-18
update aws-cdk-local to fix localstack crash

stacy-coforma:
### Description

This might be a me problem and if so just let me know, but on the `main` branch currently when I am running `./run local` my instance fails to start.  Debugging a bit, it seems to be because some of the recent dependency updates caused a mismatch between `aws-cdk-local` and some of its peer dependencies, so I've updated `aws-cdk-local` to utilize the correct versions of the updated dependencies we now have installed.

### Related ticket(s)

CMDCT-5998 probably

---
#### Localstack issue
1. Pull this branch in locally
2. (Optional) Run `./run reset` to make sure you are in a clean localstack environment
3. Run `./run local`
4. Instance should spin up without crashing

#### `jsdom` lambda issue
_This can also be verified in the ephemeral environment: https://d1worar7azgv25.cloudfront.net/_

1. Once local is running, sign in as `adminuser@test.com`
2. Update the banner in the admin page
3. Banner should be created and visible after it's saved

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/utils/sanitize/sanitize.test.ts
+++ b/services/app-api/utils/sanitize/sanitize.test.ts
@@ -156,9 +156,9 @@ describe("Test sanitizeString security", () => {
     );
   });
 
-  test("should upgrade http links to https", () => {
+  test("should allow http links", () => {
     expect(sanitizeString('<a href="http://example.com">link</a>')).toEqual(
-      '<a href="https://example.com">link</a>'
+      '<a href="http://example.com">link</a>'
     );
   });
 
--- a/services/app-api/utils/sanitize/sanitize.test.ts
+++ b/services/app-api/utils/sanitize/sanitize.test.ts
@@ -108,3 +108,112 @@ describe("Test sanitizeObject", () => {
     expect(sanitizeObject(dirtyComplexObject)).toEqual(cleanComplexObject);
   });
 });
+
+describe("Test sanitizeString security", () => {
+  test("should strip script tags and their content entirely", () => {
+    expect(sanitizeString('<script>alert("xss")</script>')).toEqual("");
+    expect(sanitizeString('<script src="evil.js"></script>')).toEqual("");
+    expect(
+      sanitizeString("<ul><script>evil()</script><li>item</li></ul>")
+    ).toEqual("<ul><li>item</li></ul>");
+  });
+
+  test("should strip event handlers from allowed tags", () => {
+    expect(sanitizeString('<a href="/" onclick="evil()">click</a>')).toEqual(
+      '<a href="/">click</a>'
+    );
+    expect(
+      sanitizeString('<strong onmouseover="evil()">text</strong>')
+    ).toEqual("<strong>text</strong>");
+    expect(sanitizeString('<em onload="evil()">text</em>')).toEqual(
+      "<em>text</em>"
+    );
+  });
+
+  test("should strip javascript:, data:, and vbscript: protocols from href", () => {
+    expect(sanitizeString('<a href="javascript:alert(1)">click</a>')).toEqual(
+      "<a>click</a>"
+    );
+    expect(
+      sanitizeString(
+        '<a href="data:text/html,<script>evil()</script>">click</a>'
+      )
+    ).toEqual("<a>click</a>");
+    expect(sanitizeString('<a href="vbscript:evil()">click</a>')).toEqual(
+      "<a>click</a>"
+    );
+  });
+
+  test("should allow safe href schemes including protocol-relative", () => {
+    expect(sanitizeString('<a href="https://example.com">link</a>')).toEqual(
+      '<a href="https://example.com">link</a>'
+    );
+    expect(
+      sanitizeString('<a href="mailto:user@example.com">link</a>')
+    ).toEqual('<a href="mailto:user@example.com">link</a>');
+    expect(sanitizeString('<a href="//example.com">link</a>')).toEqual(
+      '<a href="//example.com">link</a>'
+    );
+  });
+
+  test("should upgrade http links to https", () => {
+    expect(sanitizeString('<a href="http://example.com">link</a>')).toEqual(
+      '<a href="https://example.com">link</a>'
+    );
+  });
+
+  test("should strip style attributes from allowed tags", () => {
+    expect(
+      sanitizeString('<a href="/" style="position:fixed;top:0">link</a>')
+    ).toEqual('<a href="/">link</a>');
+    expect(sanitizeString('<strong style="color:red">text</strong>')).toEqual(
+      "<strong>text</strong>"
+    );
+  });
+
+  test("should strip disallowed block/container tags and their text content entirely", () => {
+    // Deception attack: styled overlay with fake content
+    expect(
+      sanitizeString('<div style="position:fixed;top:0">fake login</div>')
+    ).toEqual("");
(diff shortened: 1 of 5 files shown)


services/app-api/utils/sanitize/sanitize.test.ts
stacy-coforma, line 1: Copilot generated a bunch of tests and I kept the ones that proved parity between this version and the jsdom/dompurify implementation, plus the http -> https change that I put in that I am open to discussing

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
...
+  ],
+  disallowedTagsMode: "discard",
+  transformTags: {
+    a: (tagName, attribs) => {
+      // replace http with https for security
+      if (attribs.href?.startsWith("http:")) {
+        attribs = { ...attribs, href: attribs.href.replace("http:", "https:") };
+      }
+      return { tagName, attribs };
+    },
```
stacy-coforma: I added this because in my head admin users might not be thinking about http vs https, but in practice maybe we don't need this and it can be removed.

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
-import createDOMPurify from "dompurify";
-import { JSDOM } from "jsdom";
-
-const windowEmulator: any = new JSDOM("").window;
-const DOMPurify = createDOMPurify(windowEmulator);
+import sanitizeHtml, { IOptions as SanitizeHtmlOptions } from "sanitize-html";
```
stacy-coforma, line 1: since we have `sanitizeHtml` already in the PDF generation I figured we could slot it in here as well, with the stricter user-facing settings.

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
...
-  ALLOWED_ATTR: ["href", "alt"],
-  // If a tag is removed, so will all its child elements & text
-  KEEP_CONTENT: false,
-});
+const allowedTags = ["ul", "ol", "li", "a", "strong", "b", "em", "i", "u"];
+
+const sanitizeOptions: SanitizeHtmlOptions = {
+  allowedTags,
+  allowedAttributes: {
+    a: ["href", "alt"],
```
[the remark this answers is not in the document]
stacy-coforma: makes sense to me, sure

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
...
+  },
+  allowedSchemesByTag: {
+    a: ["https", "mailto"],
+  },
+  allowProtocolRelative: true,
+  // Strip the text content of disallowed tags entirely.
```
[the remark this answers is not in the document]
stacy-coforma: I want to leave this for now.  The old version explicitly allowed it (there were tests covering it) and I find it realistic that the admin might add a relative link.  Our sites are also served over `https`.

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
...
  */
-DOMPurify.setConfig({
-  // Only these tags will be allowed through
-  ALLOWED_TAGS: ["ul", "ol", "li", "a", "#text", "strong", "b", "em"],
-  // On those tags, only these attributes are allowed
-  ALLOWED_ATTR: ["href", "alt"],
-  // If a tag is removed, so will all its child elements & text
-  KEEP_CONTENT: false,
-});
+const allowedTags = ["ul", "ol", "li", "a", "strong", "b", "em"];
```
ntsummers1: Not sure how 1:1 we need to be here here, but we're missing #text from this.

services/app-api/utils/sanitize/sanitize.ts
```diff
@@ -1,30 +1,56 @@
...
+
+const sanitizeOptions: SanitizeHtmlOptions = {
+  allowedTags,
+  allowedAttributes: {
+    a: ["href"],
+  },
+  allowedSchemesByTag: {
+    a: ["https", "mailto"],
+  },
```
ntsummers1, line 20: technically, I think href and alt is allowed across _allll_ the tags that were given (Which is interesting that image isn't there). Not sure if that is also something we want to keep

services/app-api/utils/sanitize/sanitize.ts
stacy-coforma: good catch, i'd rather it be 1:1. will update

services/app-api/utils/sanitize/sanitize.ts
stacy-coforma: I feel like it's probably unlikely that our admins are going to be using alt tags, but not impossible! i'll add it back.

services/app-api/utils/sanitize/sanitize.ts
benmartin-coforma: I don't feel _too_ strongly about this, but I think I'm against it. I'm willing to bet that any domain a CMS admin would link users to, would also handle upgrading connections http -> https. Let's not take on this concern for ourselves.

stacy-coforma: word, I will remove. I was imagining scenarios in my head that I don't really think are too realistic so I'm good with removing this one.

services/app-api/utils/sanitize/sanitize.test.ts
benmartin-coforma, line 1: Musing: To some extent, these tests are verifying that our configuration says what we intend it to say. But there's a fuzzy line where it becomes verification that `sanitize-html` does what it says. The former is valuable, the latter is not. I think personally I'd have written fewer, but I won't object to merging as-is.

benmartin-coforma (approved):
Love to wave goodbye to `jsdom`.

If we're confident that this PR will remove `jsdom` from the API, and that we won't be putting it back, then we can also remove the reference to `jsdom` in deployment/constructs/lambda.ts.

For reasons I don't understand, it was necessary to specifically call out in the deployment code & explicitly install in each lambda; you'll find every app does the same. And in fact, since that file is synced via mdct-core, we may not be _able_ to remove it unless/until all our other apps do. So I guess this isn't a request - just a comment.

services/app-api/utils/sanitize/sanitize.test.ts
stacy-coforma: Yeah I just kind of let Copilot generate stuff to validate the previous version of the file primarily so I could run it after the change and make sure everything worked the same, since the settings are slightly different and I wanted to make sure it hit everything that is currently working.  The tests that were there already weren't incredibly robust.

stacy-coforma:
> Love to wave goodbye to `jsdom`.
> 
> If we're confident that this PR will remove `jsdom` from the API, and that we won't be putting it back, then we can also remove the reference to `jsdom` in deployment/constructs/lambda.ts.
> 
> For reasons I don't understand, it was necessary to specifically call out in the deployment code & explicitly install in each lambda; you'll find every app does the same. And in fact, since that file is synced via mdct-core, we may not be _able_ to remove it unless/until all our other apps do. So I guess this isn't a request - just a comment.

Yeah, I looked at that and thought about removing it in this copy but I think it'll be better to address at the source.  Maybe we can exclude this app from that particular line or something.

ntsummers1 (approved):

PR #3066, opened 2026-07-17
New Stratification Warning

sopspops:
### Description

* Add new stratification warning banner right above stratification section

### Related ticket(s)

CMDCT-6303

---
### How to test

https://d23iu8a3ddbi4m.cloudfront.net/
* Navigate to a measure with stratification setion (for example., AAB-CH)
* Fill the form and see there's a stratification banner with yellow background right above instruction under stratification section

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -47,6 +47,7 @@ export interface MeasureWrapperProps {
   detailedDescription?: string;
   year: string;
   measureId: string;
+  stratificationRequired?: CoreSetAbbr[];
   setValidationFunctions?: React.Dispatch<React.SetStateAction<any>>;
   isOtherMeasureSpecSelected?: boolean;
   isPrimaryMeasureSpecSelected?: boolean;
@@ -60,6 +61,7 @@ interface MeasureProps {
   detailedDescription?: string;
   year: string;
   measureId: string;
+  stratificationRequired?: CoreSetAbbr[];
   setValidationFunctions: Dispatch<
     SetStateAction<{
       functions: Function[];
@@ -543,6 +545,7 @@ export const MeasureWrapper = ({
                       detailedDescription={detailedDescription}
                       year={year}
                       measureId={measureId}
+                      stratificationRequired={stratificationRequired}
                       setValidationFunctions={setValidationFunctions}
                       handleSave={handleSave}
                     />
--- a/services/ui-src/src/components/MeasureWrapper/index.tsx
+++ b/services/ui-src/src/components/MeasureWrapper/index.tsx
@@ -27,7 +27,10 @@ import { CompleteCoreSets } from "./complete";
 import SharedContext from "shared/SharedContext";
 import * as Labels from "labels/Labels";
 import { coreSetBreadCrumbTitle, coreSetTitles } from "shared/coreSetByYear";
-import { featuresByYear } from "utils/featuresByYear";
+import {
+  featuresByYear,
+  isStratificationReminderBannerEnabled,
+} from "utils/featuresByYear";
 import { getStratificationBannerDescription } from "./stratificationBanner";
 import { Alert } from "@cmsgov/design-system";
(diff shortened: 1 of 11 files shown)


services/ui-src/src/shared/commonQuestions/MeasureStratification/index.tsx
```diff
@@ -216,6 +227,16 @@ export const MeasureStrat = (props: Types.OMSProps) => {
...
+      {stratificationRequired?.includes(coreSet) && (
+        <CUI.Box mt="24px" mb="32px">
+          <Alert
+            heading="Reminder: Measure Stratification Required"
+            variation="warn"
+          >
+            <CUI.Text>{stratificationBannerDescription}</CUI.Text>
+          </Alert>
+        </CUI.Box>
+      )}
```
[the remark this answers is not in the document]
stacy-coforma: this was the literal point of this ticket I believe :)

ntsummers1 (dismissed):
Looks great! Was able to verify it worked and everything acted as it should.

The following is a really small nit. Take it or leave it. Your PR works just fine :)

---- 

Theres 1 really tiny nit surrounding the rendering of the stratification reminder. 1. The stratification reminder isn't gated to just 2026. This works today because stratification isn't a pre-2026 thing (I hope I'm not mistaken on that), so 2025 forms don't show this banner. But we can do our due diligence to ensure that by adding a quick check on line 230 to make sure that the feature is meant to be there.

rocio-desantiago (approved):
Confirmed it's showing up for 2026. The banner didn't exist in 2025 and confirmed it still doesn't show there.

rocio-desantiago:
Oh quick question. What about HH set for 2025?
Screenshot 2026-07-21 at 4 58 26 PM

rocio-desantiago (changes requested):
Confirmed with Sophie. HH Core set for 2025 shouldn't be showing that notification.

services/ui-src/src/components/PrintableMeasureWrapper/index.tsx
```diff
@@ -154,6 +154,9 @@ export const PrintableMeasureWrapper = ({
   const foundMeasureDescription =
     measureDescriptions[measureData?.year]?.[measureData?.measure] ||
     measureData?.description;
+  const is2026OrLater = Number(year) >= 2026;
+  const shouldShowStratificationReminderBanner =
+    is2026OrLater && measureData?.stratificationRequired?.includes(coreSet);
```
stacy-coforma: Can we use `featuresByYear` here instead of checking for 2026 directly? May be better to create a companion one to the other existing one for visibility vs tailoring 🤔

rocio-desantiago (dismissed):
Deploy is still in progress but code makes sense 👍

services/ui-src/src/components/MeasureWrapper/index.tsx
```diff
@@ -467,6 +467,8 @@ export const MeasureWrapper = ({
     coreSet,
     featuresByYear.hasTailoredStratificationBanner
   );
+  const shouldShowStratificationReminderBanner =
+    Number(year) >= 2026 && stratificationRequired?.includes(coreSet);
```
stacy-coforma: same `featuresByYear` suggestion here!

services/ui-src/src/components/PrintableMeasureWrapper/index.tsx
benmartin-coforma: I believe `featuresByYear` would work here?

We went through a lot of effort to clean up all of the hardcoded year comparisons throughout the app and consolidate them in that file. Please make use of it whenever you can.
```suggestion
  const shouldShowStratificationReminderBanner =
    featuresByYear.hasTailoredStratificationBanner && measureData?.stratificationRequired?.includes(coreSet);
```

stacy-coforma (approved):
this works for me, I think if we later want to consolidate to one stratification flag we can do so. i like the clarity 😄

PR #3079, opened 2026-08-04
[CMDCT-6421] Replace DocRaptor with in-Lambda Prince PDF generation

tommyderami:
### Description
Switching QMR export PDF function from DocRaptor API call to using the prince binary for document generation.

### Related ticket(s)
CMDCT-6421

---
### How to test
[Deployed Env](https://d1zjggiyb62vto.cloudfront.net/)
Use the deployed cloudfront environment to fill out some measures and then test the export.  If testing locally, ensure you your machine is internet connected prior to running ```./run local``` as it will need to download the prince lambda binaries during the initial cdk deploy.

### Notes
This implementation is still missing a valid Yeslogic/Prince license file so all PDFs will have a prince comment attached to them.  Also note that the current PR downloads the x86 binaries even on an arm based M series mac because Localstack uses Rosetta emulation I believe, but there is a mac script that I will push when we switch over to Ministack.

One additional note, after approval and merge into Main, an Orca team member will need to run the pre-reqs/prince upload step in Val/Prod prior to promotion of this code to high envs.

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/deployment/stacks/api.ts
+++ b/deployment/stacks/api.ts
@@ -3,18 +3,29 @@ import {
   aws_apigateway as apigateway,
   aws_ec2 as ec2,
   aws_iam as iam,
+  aws_lambda as lambda,
   aws_logs as logs,
   aws_wafv2 as wafv2,
   CfnOutput,
   Duration,
   RemovalPolicy,
 } from "aws-cdk-lib";
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { Lambda } from "../constructs/lambda.ts";
 import { WafConstruct } from "../constructs/waf.ts";
 import { LambdaDynamoEventSource } from "../constructs/lambda-dynamo-event.ts";
 import { DynamoDBTable } from "../constructs/dynamodb-table.ts";
 import { isDefined } from "../utils/misc.ts";
 import { isLocalStack } from "../local/util.ts";
+import { PRINCE_LOCAL_DIR_REL } from "../utils/prince-asset.ts";
+
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), "../..");
+const PRINCE_GENERATED_LICENSE_PATH = join(
+  REPO_ROOT,
+  "services/app-api/bin/.generated/prince-license.dat"
+);
 
 interface CreateApiComponentsProps {
   scope: Construct;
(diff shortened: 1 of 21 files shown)


stacy-coforma:
Moved to draft due to this being on hold until other feature work is complete, to avoid prematurely introducing this change while other items are in flight

deployment/stacks/api.ts
```diff
@@ -25,8 +36,28 @@ interface CreateApiComponentsProps {
   kafkaAuthorizedSubnets: ec2.ISubnet[];
   tables: DynamoDBTable[];
   brokerString: string;
-  docraptorApiKey: string;
   kafkaClientId?: string;
+  /** Raw Prince license.dat XML; demo license used if absent. */
+  princeLicense?: string;
```
stacy-coforma, line 41: ℹ️ Based on my previous VA experience, I strongly suspected that a watermark appearing on the PDF would be unacceptable to the business owners; I checked in with Cam and he agreed that would be the case.  I understand there is some urgency around making the switch, but I think we'll need to get explicit business owner approval to merge this in before the license becomes available.  We can discuss further with Cam to get that topic in front of them.

services/ui-src/src/views/ExportAll/util.tsx
```diff
@@ -93,15 +138,38 @@ export const applyPrinceSpecificCss = (): HTMLStyleElement => {
...
+      display: block !important;
+      visibility: visible !important;
+      width: auto !important;
+      text-align: center !important;
+      text-decoration: none !important;
+      color: inherit !important;
+      background: #edf2f7 !important;
+      margin: 0 !important;
+      page-break-inside: avoid;
+    }
```
stacy-coforma, line 175: some CSS suggestions to get the TOC looking closer to how it does on main:

```suggestion
    .prince-measure-link-grid {
      display: grid !important;
      grid-template-columns: repeat(5, minmax(0, 1fr)) !important;
      column-gap: 1.25rem !important;
      row-gap: 2rem !important;
      width: 100% !important;
      max-width: 100% !important;
      margin: 0 auto 1rem auto !important;
    }
    .prince-measure-link-grid > a.chakra-button {
      display: flex !important;
      align-items: center !important;
      justify-content: center !important;
      min-height: 2.5rem !important;
      padding: 0.5rem 0.75rem !important;
      visibility: visible !important;
      width: auto !important;
      text-align: center !important;
      color: #0000e5 !important;
      background: #edf2f7 !important;
      margin: 0 !important;
      page-break-inside: avoid;
    }
```

deployment/stacks/api.ts
tommyderami: Hey I've reached back out to the folks who will provide the site license for our implementation so that we don't have to merge in without it being 'production-ready'

services/ui-src/src/views/ExportAll/util.tsx
tommyderami: I have amended my PR to include these changes--TY!

